PR #286, opened 2025-07-16
feat: add spritesheet animations to entities

BrentTB:
TODO:
- allow one entity to use animations from multiple spritesheets
- add documentation
- add tests


src/rendering/geometry/geometry.ts
```diff
@@ -33,6 +33,18 @@ export class Geometry {
     });
   }
 
+  public setTexCoords(
+    gl: WebGL2RenderingContext,
+    texCoords: Float32Array,
+  ): void {
+    const textureBuffer = this._attributes.get('a_texCoord')?.buffer;
```
stormmuller, line 40: I still don't think this is the right way to approach the problem. 

The geometry should not need to change at all. 

They way I would have approached this would be to edit the `sprite.vert.glsl` shader. I would have added 2 new instance attributes, one for the origin and one for the dimensions. Then using those values to query the sprite sheet.

https://www.youtube.com/watch?v=9VhkW9PoOdg

image

src/animations/components/image-animation-component.ts
```diff
@@ -0,0 +1,75 @@
...
+  public animationSpeedFactor: number;
+
+  // if set directly, the animation can break
+  private _currentAnimation: string;
+
+  public static readonly symbol = Symbol('ImageAnimation');
+
+  constructor(
+    entityType: string,
+    _currentAnimation: string,
```
stormmuller, line 45: remove underscore

src/animations/components/image-animation-component.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.animationIndex = animationIndex;
+    this.animationSpeedFactor = animationSpeedFactor;
+  }
+
+  public getCurrentAnimation(): string {
+    return this._currentAnimation;
+  }
+
+  public setCurrentAnimation(animation: string): void {
+    this.nextAnimationState = undefined;
```
stormmuller, line 64: never set to undefined always null

src/animations/systems/animation-manager-system.ts
```diff
@@ -0,0 +1,138 @@
...
+    animationDurationSeconds: number | number[],
+    options?: Partial<OptionalCreateAnimationSetParams>,
+  ): void {
+    const {
+      startPositionPercentage,
+      endPositionPercentage,
+      numFrames,
+      nextAnimationState,
+    } = {
+      startPositionPercentage: new Vector2(0, 0),
```
stormmuller, line 50: `Vector2.zero`

src/animations/systems/animation-manager-system.ts
```diff
@@ -0,0 +1,138 @@
...
+      endPositionPercentage: new Vector2(1, 1),
+      numFrames: spritesPerColumn * spritesPerRow,
+      ...options,
+    };
+
+    if (
+      Array.isArray(animationDurationSeconds) &&
+      animationDurationSeconds.length !== numFrames
+    ) {
+      console.error(
```
stormmuller, line 60: Throw errors

src/animations/systems/animation-manager-system.ts
```diff
@@ -0,0 +1,138 @@
...
+    const animationSet: AnimationSet = {
+      animationFrames: animationFrames,
+      numFrames,
+      nextAnimationState,
+    };
+
+    let currentAnimations = this._entityAnimationSets.get(entityType);
+
+    if (!currentAnimations) {
+      currentAnimations = new Map();
```
stormmuller, line 97: currentAnimations.clear()

src/animations/systems/animation-manager-system.ts
```diff
@@ -0,0 +1,138 @@
...
+      });
+    }
+
+    const animationSet: AnimationSet = {
+      animationFrames: animationFrames,
+      numFrames,
+      nextAnimationState,
+    };
+
+    let currentAnimations = this._entityAnimationSets.get(entityType);
```
stormmuller, line 94: `const currentAnimations = this._entityAnimationSets.get(entityType) ?? new Map();`

stormmuller (approved):

PR #293, opened 2025-07-22
Feature/particle system

BrentTB:


src/animations/systems/particle-manager-system.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    const range = (max - min) % (2 * Math.PI);
+
+    if (range === 0 && max !== min) {
+      return Math.random() * 2 * Math.PI;
+    }
+
+    return Math.random() * range + min;
+  }
```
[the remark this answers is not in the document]
BrentTB: I swap min and max if needed, so max - min will never be negative

src/animations/components/image-animation-component.ts
```diff
@@ -54,17 +55,28 @@ export class ImageAnimationComponent implements Component {
     this.currentAnimationSetName = currentAnimationSetName;
     this.animationIndex = animationIndex;
     this.animationSpeedFactor = animationSpeedFactor;
+    this.currentFrameTimeSeconds = 0;
```
stormmuller, line 58: Nice!

src/animations/components/image-animation-component.ts
```diff
@@ -54,17 +55,28 @@ export class ImageAnimationComponent implements Component {
...
+    this.nextAnimationSetName = null;
+  }
+
+  public onAnimationEnd(): void {
```
stormmuller, line 62: Components really shouldn't have any methods or logic on them. Can we move them to stand alone functions if needed.

src/animations/components/particle-component.ts
```diff
@@ -0,0 +1,73 @@
...
+      lifetimeScaleReduction,
+      height,
+      width,
+      rotation,
+      rotationSpeed,
+      lifetimeSeconds,
+      positionX,
+      positionY,
+    } = {
+      ...options,
```
stormmuller, line 46: You don't need the spread here. Looks like a copy paste from another component that had default options?

src/animations/components/particle-component.ts
```diff
@@ -0,0 +1,73 @@
+import type { Component } from '../../ecs';
+
+export interface ParticleOptions {
```
stormmuller, line 3: Should all of these really be required?

src/animations/components/particle-component.ts
```diff
@@ -0,0 +1,73 @@
...
+    this.width = width;
+    this.rotation = rotation;
+    this.rotationSpeed = rotationSpeed;
+    this.lifetimeSeconds = lifetimeSeconds;
+    this.positionX = positionX;
+    this.positionY = positionY;
+    this.ageSeconds = 0;
+  }
+
+  public update(deltaTimeInSeconds: number): void {
```
stormmuller, line 63: Logic should not exist on components. Only on systems

src/animations/components/particle-component.ts
```diff
@@ -0,0 +1,73 @@
...
+  lifetimeSeconds: number;
+  positionX: number;
+  positionY: number;
+}
+
+export class ParticleComponent implements Component {
+  public name: symbol;
+  public speed: number;
+  public originalScale: number;
+  public scale: number;
```
stormmuller, line 20: We should use the scale component.

src/animations/components/particle-component.ts
```diff
@@ -0,0 +1,73 @@
...
+  public lifetimeScaleReduction: number;
+  public height: number;
+  public width: number;
+  public rotation: number;
+  public rotationSpeed: number;
+  public lifetimeSeconds: number;
+  public positionX: number;
```
stormmuller, line 27: We should use the position component

src/animations/components/particle-component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export class ParticleComponent implements Component {
+  public name: symbol;
+  public speed: number;
+  public originalScale: number;
+  public scale: number;
+  public lifetimeScaleReduction: number;
+  public height: number;
+  public width: number;
+  public rotation: number;
```
stormmuller, line 24: We should use the rotation component

src/animations/components/particle-component.ts
```diff
@@ -0,0 +1,73 @@
...
+  public rotationSpeed: number;
+  public lifetimeSeconds: number;
```
stormmuller, line 26: We should have an age or lifetime component

src/animations/components/particle-component.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.lifetimeSeconds = lifetimeSeconds;
+    this.positionX = positionX;
+    this.positionY = positionY;
+  }
+
+  public update(deltaTimeInSeconds: number): void {
+    this.ageSeconds += deltaTimeInSeconds;
+    this.positionX += this.speed * deltaTimeInSeconds * Math.sin(this.rotation);
+    this.positionY -= this.speed * deltaTimeInSeconds * Math.cos(this.rotation);
+    const ageRatio = this.ageSeconds / this.lifetimeSeconds;
```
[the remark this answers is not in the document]
stormmuller: We should check this when creating the component, not here. But co-pilot is right, we should be checking it.

src/animations/components/particle-component.ts
```diff
@@ -0,0 +1,73 @@
...
+    this.rotationSpeed = rotationSpeed;
+    this.lifetimeSeconds = lifetimeSeconds;
+    this.positionX = positionX;
+    this.positionY = positionY;
+    this.ageSeconds = 0;
+  }
+
+  public update(deltaTimeInSeconds: number): void {
+    this.ageSeconds += deltaTimeInSeconds;
+    this.positionX += this.speed * deltaTimeInSeconds * Math.sin(this.rotation);
```
stormmuller, line 65: Using the rotation in the position calculation will make the particle rotate around come point and not around it's own center. Is this expected behaviour?

src/animations/components/particle-emitter-component.ts
```diff
@@ -0,0 +1,165 @@
+import type { Component } from '../../ecs';
+import { ForgeRenderLayer, Renderable } from '../../rendering';
+import { ParticleComponent } from './particle-component';
+
+export type positionFunction = () => number;
```
stormmuller, line 5: PositionFunction needs a better name

src/animations/components/particle-emitter-component.ts
```diff
@@ -0,0 +1,165 @@
...
+  height: 10,
+  width: 10,
+  emitDurationSeconds: 0,
+  positionX: () => 0,
+  positionY: () => 0,
+};
+
+export class ParticleEmitterComponent implements Component {
+  public name: symbol;
+  public emitters: Map<string, ParticleEmitter>;
```
stormmuller, line 44: What's the benefit of storing multiple emitters in a single component, rather each emitter being it's own entity?

src/animations/components/particle-emitter-component.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  public emitIfNotEmitting(): void {
+    if (!this.currentlyEmitting) {
+      this.startEmitting = true;
+    }
+  }
+
+  public emit(): void {
+    this.startEmitting = true;
```
stormmuller, line 163: why do we need `startEmitting` and `currentlyEmitting` can't we just use `currentlyEmitting` and set it to true when we start emitting?

src/animations/systems/animation-manager-system.ts
stormmuller, line 1: This file needs to be renamed

src/animations/systems/animation-manager-system.ts
```diff
@@ -184,20 +208,101 @@ export class SpriteAnimationManager {
...
+    for (let i = 0; i < numFrames; i++) {
+      const row = Math.floor(i / spritesPerRow);
+      const col = i % spritesPerRow;
+      animationFrames.push({
+        offset: new Vector2(
+          startPositionPercentage.x + col * spriteUVSize.x,
+          startPositionPercentage.y + row * spriteUVSize.y,
+        ),
+        scale: spriteUVSize,
+        durationSeconds: Array.isArray(animationFrameDurationSeconds)
```
stormmuller, line 258: https://forge-game-engine.github.io/Forge/docs/api/functions/enforceArray

src/animations/systems/animation-manager-system.ts
```diff
@@ -184,20 +208,101 @@ export class SpriteAnimationManager {
...
+  }
+
+  /**
+   * Sets up animation callbacks based on the provided parameters.
+   * @param animationCallbacks - An array of objects containing percentage and callback.
+   * @param numFrames - The total number of frames in the animation.
+   * @returns A Map where the key is the frame index and the value is the callback function.
+   */
+  private _setupAnimationCallbacks(
+    animationCallbacks: { percentage: number; callback: AnimationCallback }[],
```
stormmuller, line 274: No inline type definitions. Rather make a type.

src/animations/systems/animation-manager-system.ts
```diff
@@ -184,20 +208,101 @@ export class SpriteAnimationManager {
...
+      let newCallback = callback;
+
+      if (animationCallbacksMap.has(frameIndex)) {
+        const existingCallback = animationCallbacksMap.get(frameIndex);
+
+        console.debug(
+          `Multiple callbacks for frame index ${frameIndex} given. Combining callbacks into one.`,
+        );
+
+        newCallback = (entity: Entity) => {
```
stormmuller, line 297: Took me a while to understand what you're doing here 😄

Rather use forge events: https://forge-game-engine.github.io/Forge/docs/api/classes/ForgeEvent

src/animations/systems/particle-manager-system.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  private _checkStartEmitting(particleEmitter: ParticleEmitter) {
+    if (particleEmitter.startEmitting) {
+      particleEmitter.currentEmitDuration = 0;
+      particleEmitter.startEmitting = false;
+      particleEmitter.emitCount = 0;
+      particleEmitter.currentlyEmitting = true;
+      particleEmitter.amountToEmit = Math.round(
+        this._getValueInRange(particleEmitter.numParticles),
```
stormmuller, line 53: It wasn't clear to me that this was a random value

src/animations/systems/particle-manager-system.ts
```diff
@@ -0,0 +1,149 @@
...
+        particleEmitter.particles.splice(i, 1);
+
+        continue;
+      }
+
+      particle.update(this._time.deltaTimeInSeconds);
+    }
+  }
+
+  private _getValueInRange({ min, max }: MinMax): number {
```
stormmuller, line 128: https://forge-game-engine.github.io/Forge/docs/api/classes/Random

src/animations/systems/particle-manager-system.ts
```diff
@@ -0,0 +1,149 @@
...
+  public run(entity: Entity): void {
+    const particleEmitterComponent =
+      entity.getComponentRequired<ParticleEmitterComponent>(
+        ParticleEmitterComponent.symbol,
+      );
+
+    for (const particleEmitter of particleEmitterComponent.emitters.values()) {
+      particleEmitter.currentEmitDuration += this._time.deltaTimeInSeconds;
+
+      this._checkStartEmitting(particleEmitter);
```
stormmuller, line 38: I would have no idea what this function does without looking at the implementation.

src/animations/systems/particle-manager-system.ts
```diff
@@ -0,0 +1,149 @@
...
+    if (
+      !particleEmitter.currentlyEmitting ||
+      particleEmitter.emitCount >= particleEmitter.amountToEmit
+    ) {
+      particleEmitter.currentlyEmitting = false;
+
+      return;
+    }
+
+    const amountToEmit = this._getAmountToEmit(particleEmitter);
```
stormmuller, line 68: Why would I need to call `this._getAmountToEmit(particleEmitter)` when `particleEmitter.amountToEmit` exists?

src/animations/systems/particle-manager-system.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    for (let i = 0; i < amountToEmit; i++) {
```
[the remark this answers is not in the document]
stormmuller: 100% https://forge-game-engine.github.io/Forge/docs/api/classes/ObjectPool

src/animations/systems/particle-manager-system.ts
```diff
@@ -0,0 +1,149 @@
...
+      const speed = this._getValueInRange(particleEmitter.speed);
```
stormmuller, line 71: I just realised `speed`, `scale`, `lifetime`, etc. aren't numbers just ranges. 

The names are misleading.

src/animations/systems/particle-manager-system.ts
```diff
@@ -0,0 +1,149 @@
...
+      const lifetimeSeconds = this._getValueInRange(particleEmitter.lifetime);
+
+      const rotation = this._getValueInRangeRadians(particleEmitter.rotation);
+
+      const rotationSpeed = this._getValueInRange(
+        particleEmitter.rotationSpeed,
+      );
+
+      particleEmitter.particles.push(
+        new ParticleComponent({
```
stormmuller, line 84: Components should only be added to entities, they are being abused when using them like this.

src/animations/systems/particle-manager-system.ts
```diff
@@ -0,0 +1,149 @@
...
+          speed,
+          originalScale,
+          lifetimeSeconds,
+          rotation,
+          rotationSpeed,
+
+          lifetimeScaleReduction: particleEmitter.lifetimeScaleReduction,
+          height: particleEmitter.height,
+          width: particleEmitter.width,
+          positionX: particleEmitter.positionX(),
```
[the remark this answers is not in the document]
BrentTB: to allow particles to spwan from an area, instead of a single spot. For example, the jump particles spawn from a horizontal line (instead of a point) using these functions. It also lets the spawn position change over time (like the attacking particle, whose spawn point moves during the attack animation)

src/animations/systems/particle-manager-system.ts
BrentTB: fixed

src/animations/components/particle-emitter-component.ts
BrentTB: This was done so that multiple different emitters can exist on the same entity. 
For example, in the demo game (`animationDemo.ts`), this lets the animation callback functions to create jump / attack particles. (eg: as used in the `createAdventurerControllableAnimationSets` function

src/animations/components/particle-emitter-component.ts
BrentTB: startEmitting is used to set some things before any particles are created. For example, in `particle-emitter-system.ts`, in the `_checkStartEmitting` function. 

This decides the amount to emit, sets the emit duration, and a few others. But I dont want this to run every iteration, only once per particle emission

src/animations/components/particle-emitter-component.ts
```diff
@@ -0,0 +1,165 @@
+import type { Component } from '../../ecs';
+import { ForgeRenderLayer, Renderable } from '../../rendering';
+import { ParticleComponent } from './particle-component';
+
+export type positionFunction = () => number;
+
+export interface MinMax {
+  min: number;
```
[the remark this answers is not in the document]
BrentTB: currently, this validation is done in the `_getRandomValueInRange` and `_getRandomValueInRangeRadians` functions (in `particle-emitter-system.ts`. In there, it swaps the min and max value, if min > max

src/animations/systems/particle-emitter-system.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  private _getValueInRangeRadians({ min, max }: MinMax): number {
+    if (min > max) {
+      [min, max] = [max, min];
+    }
+
+    const range = (max - min) % 360;
+
+    if (range === 0 && max !== min) {
+      return Math.random() * 360;
```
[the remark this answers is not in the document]
BrentTB: if the range is a multiple of 360, then the user wants any angle in a full circle. So we dont have to add min

src/animations/systems/animation-manager-system.ts
BrentTB: Done 👍

src/animations/components/particle-component.ts
BrentTB: The particles just needed the ability to rotate (so that they did not always move in a straight line), so that is fine here

src/particles/systems/particle-emitter-system.ts
```diff
@@ -0,0 +1,159 @@
...
+    return this._random.randomFloat(min, max);
+  }
+
+  private _getRandomValueInRangeDegrees({ min, max }: MinMaxRange): number {
+    if (min > max) {
+      [min, max] = [max, min];
+    }
+
+    const range = (max - min) % 360;
+
```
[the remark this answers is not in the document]
BrentTB: we swap min and max if min > max, so this cannot happen

src/animations/systems/animation-set-manager.ts
```diff
@@ -173,31 +203,106 @@ export class SpriteAnimationManager {
...
   /**
-   * Retrieves the current animation frame for a given ImageAnimationComponent.
-   * @param imageAnimationComponent - The ImageAnimationComponent to get the current frame for.
+   * Retrieves the current animation frame for a given SpriteAnimationComponent.
+   * @param spriteAnimationComponent - The SpriteAnimationComponent to get the current frame for.
    * @returns The current AnimationFrame or null if the component is not provided.
    */
   public getAnimationFrame(
-    imageAnimationComponent: ImageAnimationComponent | null,
+    spriteAnimationComponent: SpriteAnimationComponent | null,
```
stormmuller, line 211: Why can this be null?

src/animations/systems/animation-set-manager.ts
```diff
@@ -104,12 +137,11 @@ export class SpriteAnimationManager {
...
       endPositionPercentage,
       numFrames,
-      nextAnimationSetName: nextAnimationSetName,
+      nextAnimationSetName,
+      animationCallbacks,
     } = {
-      startPositionPercentage: Vector2.zero,
-      endPositionPercentage: Vector2.one,
+      ...defaultCreateAnimationSetParams,
       numFrames: spritesPerColumn * spritesPerRow,
```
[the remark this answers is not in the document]
BrentTB: yup, the thought being if the animations do not fully fit the grid. ie: if it was a 4x4 grid of frames, but the last 2 were empty (so there are only actually 12 animation frames) we can still use this properly

src/animations/systems/animation-set-manager.ts
```diff
@@ -62,14 +70,39 @@ export interface OptionalCreateAnimationSetParams {
...
    */
   nextAnimationSetName: string | null;
+
+  /**
+   * Callback functions to run at specific percentage points of the animation.
+   * percentage is a number between 0 and 1, representing the progress of the animation.
+   * percentage = 0 means the first frame of the animation, and percentage = 1 means the last frame of the animation.
+   * @default []
+   */
+  animationCallbacks: Array<{
```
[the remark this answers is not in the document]
BrentTB: I had an array so that the same index could refer to multiple events.
But I was getting a lot of problems doing this, so I decided that the user will just enter the event map themself. And then they dont have to enter percentages and we calculate the frames, they just enter the frame they want the event to be fired on

src/animations/sprite-animation-helper.ts
```diff
@@ -0,0 +1,46 @@
...
+  spriteAnimationComponent: SpriteAnimationComponent,
+): void {
+  if (!spriteAnimationComponent.nextAnimationSetName) {
+    throw new Error(
+      `No next animation set name specified for entity type "${spriteAnimationComponent.entityType}" with current animation "${spriteAnimationComponent.currentAnimationSetName}". ` +
+        'Set the nextAnimationSetName property on spriteAnimationComponent component instance, or use setCurrentAnimation() to set the next animation.',
+    );
+  }
+
+  setCurrentAnimation(
```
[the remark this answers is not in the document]
BrentTB: I wanted to separate going to the next animation from setting the next animation. 
For example, when the user presses 'd' right before a jump animation ends, we can then have the run animation go after the end of the jump. But when we swap to that run animation, we wont necessarily have the next animation that will come after that

Thus we can go to the next animation, without resetting the animation that will come after that

PR #347, opened 2025-10-18
Module boundry

stormmuller:


src/rendering/shaders/utils/create-texture-from-image.ts
```diff
@@ -8,7 +8,7 @@
 export const createTextureFromImage = (
   gl: WebGL2RenderingContext,
   image: TexImageSource,
-) => {
+): WebGLTexture | null => {
```
stormmuller: @copilot make sure this function never returns null (do a null check + error)

PR #485, opened 2026-07-14
Bloom

stormmuller:


src/physics/systems/physics.system.ts
```diff
@@ -25,7 +25,15 @@ export const createPhysicsEcsSystem = (
...
-  const registeredEntities = new Map<RigidBody, number>();
+  // Keyed by entity rather than by `RigidBody` instance: `EcsWorld` recycles
+  // entity ids as soon as an entity is removed, so a `PhysicsBodyId`
+  // component removed and re-added under the same entity id within a
+  // single tick (e.g. an object pool or a "respawn" that removes a batch
+  // of entities and immediately creates a new one) would otherwise be
+  // indistinguishable from that entity never having changed bodies at
+  // all, leaving the old `RigidBody` registered in `physicsWorld` forever
+  // as an invisible, un-synced collider.
+  const registeredEntities = new Map<number, RigidBody>();
```
stormmuller, line 36: Is this the best solution?

src/physics/systems/physics.system.ts
stormmuller: Logically, this will never be false. The query includes physics body.

src/physics/systems/physics.system.ts
```diff
@@ -45,18 +53,25 @@ export const createPhysicsEcsSystem = (
         }
 
         const { physicsBody } = physicsBodyComponent;
+        const registeredBody = registeredEntities.get(entity);
 
-        if (!registeredEntities.has(physicsBody)) {
-          physicsWorld.addBody(physicsBody);
-          registeredEntities.set(physicsBody, entity);
-          physicsBody.userData = entity;
+        if (registeredBody === physicsBody) {
```
stormmuller: Ok so if it's already been registered on a previous frame, don't register it again?

src/physics/systems/physics.system.ts
```diff
@@ -45,18 +53,25 @@ export const createPhysicsEcsSystem = (
...
+          continue;
+        }
+
+        if (registeredBody) {
```
stormmuller: If they aren't the same body, then remove the registered body from the physics world, it is stale as there is a new body registered to this entityId? This will only every happen because the entityId was recycled.

src/physics/systems/physics.system.ts
```diff
@@ -45,18 +53,25 @@ export const createPhysicsEcsSystem = (
...
+          physicsWorld.removeBody(registeredBody);
         }
+
+        physicsWorld.addBody(physicsBody);
```
stormmuller, line 63: Register the new body (or existing body, I assume this call is idempotent?)

src/physics/systems/physics.system.ts
```diff
@@ -45,18 +53,25 @@ export const createPhysicsEcsSystem = (
...
+        registeredEntities.set(entity, physicsBody);
```
stormmuller, line 64: register the new (or existing) body. calling `set` is idempotent.

src/physics/systems/physics.system.ts
```diff
@@ -45,18 +53,25 @@ export const createPhysicsEcsSystem = (
...
+        physicsBody.userData = entity;
       }
 
-      for (const [physicsBody, entity] of registeredEntities) {
+      for (const [entity, physicsBody] of registeredEntities) {
```
stormmuller: Another loop? This seems like a clean up step in the case that the entity is not part of the query result.

src/physics/systems/physics.system.ts
stormmuller, line 68: Why are we running the physics step in the before hook? Seems more like an after hook concern.

stormmuller, line 79: Why are they different? The physics body needs to be in "render" space.

stormmuller: Ok I see, we run the physics simulation and then for each queried entity we update the position and rotation components.

stormmuller, line 75: Should the processing for kinematic bodies happen before of after the simulation runs?

src/physics/systems/physics.system.ts
```diff
@@ -82,12 +97,13 @@ export const createPhysicsEcsSystem = (
       }
     },
     cleanupEntities: (result) => {
-      const [physicsBodyComponent] = result.components;
```
stormmuller, line 85: Another clean up step? Does this actually remove anything new?

src/physics/systems/physics.system.ts
```diff
@@ -82,12 +97,13 @@ export const createPhysicsEcsSystem = (
       }
     },
     cleanupEntities: (result) => {
```
stormmuller, line 89: We should rename this hook

src/rendering/enums/render-target-format.enum.ts
```diff
@@ -0,0 +1,27 @@
...
+   * moment a fragment shader writes them. The default for every render
+   * target.
+   */
+  ldr: 'ldr',
+
+  /**
+   * Half-float (`RGBA16F`) color storage. Values above `1` survive
+   * intermediate passes instead of being clamped, so bloom and tone mapping
+   * can operate on true HDR brightness. Requires the
+   * `EXT_color_buffer_float` WebGL2 extension; silently falls back to `ldr`
```
stormmuller, line 23: Is the fallback logged?

src/rendering/utilities/create-image-sprite.ts
```diff
@@ -7,23 +7,83 @@ import {
...
  * @param frameDimensions - The dimensions of a single frame in the image (for sprite sheets).
+ * @param emissiveMapOptions - An emissive map to add on top of the sprite's tinted albedo, unaffected by lighting. Omit for a sprite with no emissive contribution.
  * @returns The created sprite.
  */
 export function createImageSprite(
   image: HTMLImageElement,
   renderContext: RenderContext,
   layer: number,
   frameDimensions: Vector2 = new Vector2(image.width, image.height),
+  emissiveMapOptions?: EmissiveMapOptions,
```
stormmuller: This pattern isn't scalable when there are other Maps/Options

Issue #471, opened 2026-06-26
Add support for n-slicing sprites


stormmuller:
### The user should be able to slice sprites

The user can define multiple "slices" and for each region created, they should be able to define how it scales (stretch or tile)

- [ ] User defines slices in the [sprite creation options](https://forge-game-engine.github.io/Forge/docs/api/type-aliases/SpriteOptions)
- [ ] User defines how regions behave when scaled (stretch by default)
- [ ] The sprite shader and renderer should be updated to ensure that the regions render as expected

PR #504, opened 2026-07-17
feat(rendering): add nine-slice sprite support

Nitjsefnie:
Implements #471 as classic 9-slice (border insets, the Unity/CSS/Godot convention), stretch-mode v1.

## The seam, and one deliberate deviation from the issue's pointer

The issue points at `SpriteOptions` in `src/rendering/sprite.ts` — but on `dev`, that `Sprite` class is **particle-only** (its sole consumer is the particle emitter) and never reaches the renderer. The real render path is `SpriteEcsComponent` → one `RenderCommand` per sprite → instance-buffer fill → one instanced draw. So slicing lands there: **a sliced sprite emits one render command per non-empty slice** (up to 9), and every sub-quad flows through the existing sort/batch/buffer/draw path — zero changes to `includeBatch`, the draw call, `floatsPerInstance`, or any GLSL. Each sub-quad's placement is encoded by solving the shader's own pivot algebra, so rotation/scale/tint stay shared and correct for free.

## API

`slices?: { left, right, top, bottom }` (texture pixels) + `sliceMode?: 'stretch'` (literal union — `'tile'` becomes a non-breaking addition later) on the sprite component and `createImageSprite` (which auto-fills the new `textureDimensions` from the frame). Construction-time validation: non-negative insets, left+right < width, top+bottom < height. No `slices` → exactly the previous single-quad path (never calls the slice code); zero insets on an axis collapse cleanly instead of emitting degenerate quads.

## Design calls explicitly yours to overrule

1. Fields on the ECS component (+ `createImageSprite`), not the particle-only `Sprite` class — the biggest deviation from the issue's pointer, forced by the render path.
2. The new `textureDimensions` field (insets need a pixel reference); could instead fold into the `slices` object.
3. Corner convention: 1 texture pixel = 1 world unit (corners keep native size at any sprite scale); alternative is scaling corners by `width/textureWidth`.
4. Insets measure against the sampled frame, not the full atlas image.
5. `sliceMode` stored but not yet consumed ('stretch' is the only behavior).

Known v1 limits, documented in code: per-frame allocation of the 9 sub-instances (matches the existing per-entity-per-frame style; poolable later), and negative-scale flips mirror geometry but not per-cell UVs.

## Verification

All CI jobs run locally and green: ESLint (only pre-existing todo-tag warnings), tsc (tsconfig.ci), cspell, check-exports (node16 + bundler 🟢), build, **Vitest 936 passed / 130 files** (43 tests in the touched files), commitlint. The slice math is a pure function with direct tests. Mutation-checked (implementer, then independently re-run before opening): swapping geometry-edge insets, swapping UV-edge insets, flipping the pivot sign, and collapsing a sliced sprite to one emitted quad each fail exactly the tests that pin them — the last one re-verified by hand before this PR.

---
🤖 This PR was written with AI assistance (Claude), directed and verified by a human-supervised workflow.

### DIFF
--- a/src/rendering/components/sprite-component.test.ts
+++ b/src/rendering/components/sprite-component.test.ts
@@ -71,6 +71,54 @@ describe('addSpriteComponent', () => {
     expect(world.getComponent(entity, spriteId)).toBe(component);
   });
 
+  it('attaches nine-slice options when given valid insets and texture dimensions', () => {
+    const world = new EcsWorld();
+    const entity = world.createEntity();
+    const renderable = createRenderable();
+
+    addSpriteComponent(world, entity, {
+      width: 100,
+      height: 100,
+      renderable,
+      slices: { left: 16, right: 16, top: 16, bottom: 16 },
+      textureDimensions: new Vector2(64, 64),
+    });
+
+    expect(world.getComponent(entity, spriteId)).toMatchObject({
+      slices: { left: 16, right: 16, top: 16, bottom: 16 },
+      textureDimensions: new Vector2(64, 64),
+    });
+  });
+
+  it('throws when slices are given without texture dimensions', () => {
+    const world = new EcsWorld();
+    const entity = world.createEntity();
+
+    expect(() =>
+      addSpriteComponent(world, entity, {
+        width: 100,
+        height: 100,
+        renderable: createRenderable(),
+        slices: { left: 16, right: 16, top: 16, bottom: 16 },
+      }),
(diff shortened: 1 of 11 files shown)


stormmuller (changes requested):
Please write the description yourself. There is too much noise in the description that is not relevant to the change at all.

You used Claude, the Claude.md file states that the documentation site needs to be updated along with any change. That includes a demo. 

If I get an AI response to this comment, I'm closing the PR.

PR #507, opened 2026-07-19
Add torque application and angular velocity motor ECS systems

### LINKED ISSUE #500: Add Torque and Angular Velocity Motor component
If you want to apply pure rotational torque to any rigid body (a spinning coin, a thruster-controlled spaceship, a fan, or a car chassis), you only need a transient component to represent the torque force during the current physics tick (i.e. a Torque component):

```
interface AppliedTorque {
    float value; // Torque in N*m to apply this frame
};
```

For a controlled motor (which targets a specific velocity and has a torque limit) you need a Angular Velocity Motor

```
interface AngularVelocityMotor {
    float target_velocity; // Rad/s
    float max_torque;      // Max torque capacity (N*m)
};
```

stormmuller:
## Summary

Adds support for applying torque to rigid bodies and controlling angular velocity through ECS components. This enables continuous rotation control (like spinning fans, thrusters, or car wheels) with both direct torque application and target velocity-driven motor components.

## Key Changes

- **`RigidBody.applyTorque(torque, deltaTimeInSeconds)`**: New method that changes angular velocity by `torque * inverseInertia * deltaTimeInSeconds`, mirroring how gravity works for linear motion. Has no effect on static bodies.

- **`AppliedTorqueEcsComponent` and system**: ECS component that applies a torque value to a physics body each tick, then resets to zero. Enables one-shot torque kicks or continuous spin when the value is re-applied every frame.

- **`AngularVelocityMotorEcsComponent` and system**: ECS component that drives a body toward a target angular velocity, spending no more than a specified maximum torque per tick. Useful for holding a rotation speed (fan at rated RPM, wheel matching throttle) rather than managing torque manually.

- **Documentation updates**: Expanded the "Applying Forces" guide with a new "Torque: spinning a body" section covering `applyTorque`, both ECS components, usage examples, and registration order requirements.

- **Test coverage**: Added comprehensive tests for both the `applyTorque` method and both ECS systems, covering normal operation, torque clamping, static body handling, and component attachment.

## Implementation Details

- Both ECS systems must be registered before `createPhysicsEcsSystem` to ensure torque is applied in the same tick as physics integration.
- The angular velocity motor uses a responsiveness calculation (`inverseInertia * deltaTimeInSeconds`) to compute the required torque, then clamps it to the maximum allowed.
- Applied torque component resets its value after each application, preventing unintended repeated application across ticks.
- Updated component and system index exports to expose the new APIs.

https://claude.ai/code/session_01EdvrBcYntaZBjk69pSdevN


stormmuller:
Please add an option for angular drag, defaulted to `0` and ensure that the "left flywheel" demo has some drag applied.

src/physics/systems/applied-torque.system.ts
```diff
@@ -0,0 +1,36 @@
...
+ * `RigidBody` every tick (via `RigidBody.applyTorque`), then resets `value`
+ * back to `0`.
+ *
+ * Must be registered with the `EcsWorld` before `createPhysicsEcsSystem` (or
+ * with an earlier `registrationOrder`), so torque applied this tick is
+ * reflected in the same tick's `physicsWorld.step`.
+ * @param time - The time instance used to scale torque by the tick's delta
+ * time.
+ */
+export const createAppliedTorqueEcsSystem = (
```
stormmuller: I don't think the `TorqueEcsSystem` and `AppliedTorqueEcsComponent` make sense to belong in the engine. In actual game logic the game would have some system that queries for the RB and applies torque directly. Remove this system and the AppliedTorque component. And update the demo accordingly. In the demo we can introduce a demo specific tag like `MovableFlyWheelEcsComponent` or similar to act as a discriminator.

src/physics/systems/angular-velocity-motor.system.ts
```diff
@@ -0,0 +1,56 @@
...
+ * and the tick's delta time, then clamped to `[-maxTorque, maxTorque]`
+ * before being applied via `RigidBody.applyTorque`.
+ *
+ * Must be registered with the `EcsWorld` before `createPhysicsEcsSystem` (or
+ * with an earlier `registrationOrder`), so the motor's torque is reflected
+ * in the same tick's `physicsWorld.step`.
+ * @param time - The time instance used to scale torque by the tick's delta
+ * time.
+ */
+export const createAngularVelocityMotorEcsSystem = (
```
stormmuller, line 26: Do we really need a new system to handle Motors specifically? Could we rather include it in the physics system? 

If it's more idiomatic to keep it as is then let me know and don't make a change.

src/physics/systems/applied-torque.system.ts
stormmuller: Agreed, done in 766f351. Removed `AppliedTorqueEcsComponent`/`createAppliedTorqueEcsSystem` entirely (kept `AngularVelocityMotorEcsComponent`, since its target-velocity/max-torque logic is non-trivial and broadly reusable, unlike the removed one which was just a thin "set value, apply, reset to 0" wrapper).

For the demo, I kept `ThrusterEcsComponent` rather than adding a separate bare tag: it already carries the `holdAction`/`torque` data *and* acts as the discriminator in `createThrusterEcsSystem`'s query (`[thrusterId, PhysicsBodyId]`), which now calls `physicsBody.applyTorque` directly - the same pattern the demo's `GustEcsComponent`/`createGustEcsSystem` already used for the motor's perturbation. Happy to split out a separate marker tag if you'd rather keep "identifies this entity" and "carries its data" as two distinct components, but it seemed like unnecessary indirection here since nothing else needs to query for "any flywheel" without also needing the thruster-specific data.

---

src/physics/systems/angular-velocity-motor.system.ts
stormmuller: I'd lean towards keeping it as its own system, for consistency with the precedent `createRevoluteJointEcsSystem`/`createPrismaticJointEcsSystem` already set: both are separate systems that have to run *before* `createPhysicsEcsSystem` in the same tick (documented as a "registration order" caution in the docs), rather than being folded into it, even though they're just as tightly coupled to "affect the body before the step" as the motor is.

Beyond precedent, `createPhysicsEcsSystem`'s job is specifically registering bodies, stepping `physicsWorld`, and syncing ECS position/rotation - it queries `[PhysicsBodyId, positionId, rotationId]`. Folding the motor in would mean it also has to query `AngularVelocityMotorEcsComponent`, conflating two independent concerns (a body with a motor doesn't need position/rotation synced any differently), and would force every user to pay for the motor's iteration even if they never use it. Keeping it separate lets you only register it when a game actually has motorized entities, same as the joint systems.

So: I think it's more idiomatic as-is, but let me know if you'd rather fold it in anyway and I'll make the change.

---

stormmuller: Noted, and agreed. 

Let's rename the physics system. Because at it stands, it sounds like it handle all physics behaviours when in reality it just syncs the physics world with the ECS world and vice-versa.

stormmuller: Done in 53d7163: renamed `createPhysicsEcsSystem` → `createPhysicsSyncEcsSystem` (file `physics.system.ts` → `physics-sync.system.ts`), and updated every call site across the engine, docs, and demos to match. Kept it as its own system per the reasoning above.

---

PR #528, opened 2026-07-23
fix(physics): stop rigid bodies settling deeply embedded in dense piles

stormmuller:
## Summary

Bodies at the bottom of a pile/stack (see the `physics` demo) were settling visibly embedded in each other instead of resting side-by-side. Root cause was two compounding bugs in `resolveCollision`:

- **Under-strength impulses at multi-contact manifolds**: normal and friction impulse magnitudes were divided by the manifold's contact point count. For 2-point polygon-polygon manifolds this halved every impulse, so tilted/rocking contacts never fully converged in one pass and could snowball into severe interpenetration under load.
- **A positional-correction cap too small for the shapes involved**: the per-step correction budget was a flat `2` units, far smaller than typical shape sizes (tens of units in the demo). Once a chaotic pile compressed a pair past that distance in a single step (common with many bodies free-falling into a container at once), correction could never catch back up — the pair froze in a deep, stable overlap indefinitely, worst at the bottom of the pile where load (and thus initial compression) is highest.

## Changes

- Removed the erroneous `/ contactPoints.length` division from both the normal and friction impulse calculations in `resolveCollision`.
- Replaced the flat `MAX_LINEAR_CORRECTION_PER_STEP = 2` cap with a per-manifold cap scaled to the smaller body's bounding radius (`MAX_LINEAR_CORRECTION_FRACTION = 1.0`), so the cap stays meaningful regardless of a game's shape scale.
- Updated a stale doc reference in `CollisionManifold.correctionApplied` to the renamed constant.
- Added regression tests: full velocity cancellation across multiple contact points in one call, and the correction cap scaling with body size.
- Added a Keep a Changelog entry under `[Unreleased]`.

## Verification

- Reproduced the bug with a scripted 120-shape pile matching the demo's parameters: worst-case penetration depth was ~40-48px and stayed there for 2+ minutes of simulated time (not a slow transient — a genuine stuck state). After the fix, worst-case depth dropped to ~15-20px, and the count of low-velocity "resting" pairs still overlapping by >5px fell roughly 4x.
- Confirmed visually: rebuilt `dist`, ran the docs site, and screenshotted the `physics` demo before/after the fix — the before shot reproduces the reported bug (shapes visibly melted into each other at the bottom of the pile); the after shot shows clearly separated shapes.
- Checked for regressions: a simple resting stack, a single box on the ground, and a hard-impact drop onto a settled stack all behave the same as before (no new jitter/vibration introduced by the larger correction cap).
- `npm run check-types`, `npm test` (1045 tests), `npm run lint` (0 errors, pre-existing warnings only), `npm run cspell`, and `npm run check-exports` all pass.
- `documentation-site`: `npm run build` and `npm run typecheck` pass; no demo source changes were needed since `resolveCollision`'s public signature is unchanged.

## Test plan

- [x] `npm run check-types`
- [x] `npm test`
- [x] `npm run lint`
- [x] `npm run cspell`
- [x] `npm run check-exports`
- [x] `documentation-site` build + typecheck
- [x] Visual check of the `physics` demo before/after

---
_Generated by [Claude Code](https://claude.ai/code/session_01T42uEAwc1rPfwvJeuYvVPq)_


stormmuller:
Follow-up commit: after review the shapes in the physics demo were visibly vibrating. Root cause was a side effect of the first fix here — raising the positional-correction cap to fix the embedding bug also let it compound across all 8 velocity iterations (~83% of a manifold's depth per step), which was large enough to overshoot a body out of one overlapping neighbor and into another, oscillating frame to frame. That compounding is exactly what the original (too-small) cap had been suppressing.

Fix: positional correction now runs once per step (on the last velocity iteration, after velocity has settled) instead of on every iteration, and `CORRECTION_PERCENT` was raised from 0.2 to 0.5 to keep convergence speed now that it's no longer compounding.

Verified with a scripted 120-shape pile: worst-case per-step displacement of resting bodies dropped to well under a pixel (vs. ~7.5px with the cap-only fix, and ~2px on the pre-fix baseline), and confirmed 0% pixel difference between browser frames of the settled demo 200ms apart. Two existing tests had hardcoded expected positions baked in from the old compounding behavior — updated those to match.


---

stormmuller:
Merged `dev` in again to pick up the v0.23.1 release that happened while this PR was open. The automatic merge put this PR's changelog bullet under the new `## [0.23.1]` header instead of leaving it under `Unreleased` (git's line-based merge doesn't know that section is historical) — that release happened before this PR merged, so the physics fix was never actually part of it. Moved the bullet back to `Unreleased` in a follow-up commit; the `0.23.1` section now matches `dev` exactly.


---

stormmuller:
Closed, as I didn't see a difference between the current state and this fix.

PR #530, opened 2026-07-24
docs(demos): nine-slice stretched sprites, TerrainShape hill-climb course, joint arm graphics

stormmuller:
## Summary

- **Nine-slice stretched sprites**: converted every demo sprite that was being naively, non-uniformly stretched to use `NineSliceOptions` so rounded corners/caps stay crisp instead of smearing — torque's flywheel bars, revolute-joint's door, linear-spring-damper's spring line, brick-breaker's paddle, prismatic-joint's sliders, and Newton's Cradle's frame (which also fixes a pre-existing bug where the frame ignored its computed width entirely).
- **Hill Climb Racer terrain**: replaced the stepped `PolygonShape` column terrain with a single `TerrainShape` ground body built from a smooth curve (`buildTerrainCurve` + `createTerrainMesh`), matching the Rolling Ball demo's approach.
- **Newton's Cradle / Wrecking Ball connection graphics**: added a demo-only `ArmLineEcsComponent`/system pair that draws a nine-sliced rod sprite from each `RevoluteJoint`'s pivot to its ball every tick, since the joint itself has no visual representation.
- **Nine-slice demo fix**: the panel breathing animation only ever changed `sprite.width` (height was commented out), and the UI described 3 comparison panels while only 2 were built — both fixed.

## Bug found and fixed (demo-only, not an engine bug)

While building the new terrain, the course rendered fine but the car fell straight through it. Root cause: `TerrainShape`'s required 180° body rotation (to make its solid slab face "down") negates **both** x and y — so control points authored directly as `Vector2(x, heightAt(x))` land at world `(-x, -heightAt(x))` once rotated. That mirrored the whole course and left almost no terrain ahead of the car's spawn point, so the front wheel started off the edge and rolled into the void. Fixed by negating both coordinates when building control points (and reversing the array, since negating x reverses the required increasing-x order) so world space matches `heightAt`'s coordinates directly. Added a caution note to the terrain docs (`documentation-site/docs/docs/physics/terrain.md`) so the next person authoring a one-directional (non-bounded) `TerrainShape` course doesn't hit the same thing.

No engine (`/src`) code changed — this whole PR is demo-only, so no `CHANGELOG.md` entry per the repo's changelog rules.

## Test plan

- [x] `npm run check-types` / `npm test` / `npm run lint` / `npm run cspell` / `npm run check-exports` — all pass (unchanged, since no `/src` changed)
- [x] `npm run build` (root) + `documentation-site`: `npm run typecheck` and `npm run build` — all pass
- [x] Loaded every touched demo in a real browser (Playwright + headless Chromium) and visually verified: Hill Climb Racer (car spawns, sits stably on the new terrain, drives and climbs correctly under input), Newton's Cradle (visible arms + resized frame), Wrecking Ball (visible arm), Nine-Slice (3 panels, naive vs sliced distortion visible), Torque, Revolute Joint, Prismatic Joint, Linear Spring/Damper, Brick Breaker (all show rounded/undistorted nine-sliced sprites)

---
_Generated by [Claude Code](https://claude.ai/code/session_01HMW3rFRURenXGrzGVf433F)_


stormmuller:
Closed as there have been a lot of changes that would influence this solution and we need to start fresh

PR #531, opened 2026-07-25
docs(demos): design demos for a fixed 1920x1080 canvas

stormmuller:
## Summary

- Documentation-site demos previously computed entity positions, spawn bounds, and layout math as ratios of `renderContext.canvas.width` / `renderContext.canvas.height`, so a demo's layout scaled with whatever container size it happened to render at.
- Replaced those ratio calculations with absolute pixel values authored against a fixed 1920x1080 canvas, across 12 demos (torque, revolute-joint, prismatic-joint, linear-spring-damper, physics, stress-test, easing-functions, nine-slice, newtons-cradle, brick-breaker, particles, space-shooter).
- Demos are intentionally no longer expected to look correct at other canvas resolutions.
- Left two categories of canvas-size reads untouched, since they're not "entity in the world" placement: mouse/pointer-to-world coordinate conversion (`screenToWorldSpace`, needs the live canvas size to map real cursor positions), and GPU render-target sizing for the space-shooter demo's bloom/blur post-processing (must match the actual rendered surface).
- Simplified `nine-slice`'s panel spacing, which previously clamped itself to fit a "narrow container" — no longer needed now that the canvas size is a fixed assumption.
- Dropped the now-unused `renderContext` parameter from `spawnPlayer` and `createAsteroidEcsSystem` in the space-shooter demo, since their only use of it was the canvas-size read that's now a constant.
- Updated `AGENTS.md`'s "Documentation Site Demos" section to document the 1920x1080 convention and its two exceptions for future changes.

## Test plan

- [x] `npm run check-types`, `npm test`, `npm run lint`, `npm run cspell`, `npm run check-exports` all pass at the repo root
- [x] `npm run build` (root) to refresh `/dist`
- [x] `documentation-site`: `npm run typecheck` and `npm run build` pass
- [x] Started the docs site locally and loaded all 12 changed demos with the canvas forced to a genuine 1920x1080 size — confirmed entities render in their intended positions with no console errors

---
_Generated by [Claude Code](https://claude.ai/code/session_01FG27RSJkn29E1cA6UnmhDk)_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -422,6 +422,20 @@ import it, e.g. `grep -rl "/physics" documentation-site/src/pages/demos`):
 See also step 9 of `CLAUDE.md`'s verification checklist, which makes this
 mandatory before marking a task complete.
 
+**Demos assume a fixed 1920x1080 canvas.** Entity positions, spawn bounds,
+and layout math in demo code (e.g. `_create-boundaries.ts`,
+`_create-asteroids.ts`) use absolute pixel values authored against a
+1920x1080 world, not ratios derived from `renderContext.canvas.width` /
+`renderContext.canvas.height` or the live container size. Demos are not
+expected to look correct at any other canvas size. The canvas itself is
+still sized dynamically from its container (`createCanvas` /
+`RenderContext` are unchanged); only the entities placed inside it assume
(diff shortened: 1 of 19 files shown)


stormmuller:
The real issue here stemmed from the lack of a "world unit" in the engine. This was fixed in #533

PR #532, opened 2026-07-25
test(e2e): scaffold real-browser Playwright suite for camera pan/zoom

stormmuller:
## Summary

Scaffolds a real-browser integration/e2e test suite under `e2e/`: independent of `/demo` and `/documentation-site`, testing directly against `/src` as if it were the only thing in the repo.

- New `e2e/fixtures/` Vite app (served via new `vite.config.e2e.js`, matching the `vite.config.demo.js` pattern): a single HTML shell + `harness.ts` that loads a named scene (`?scene=`) via `import.meta.glob` and exposes its handle on `window.__forgeTestHooks`. A failed scene renders its error into the page and logs it, instead of leaving an indistinguishable blank canvas.
- First scene, `e2e/fixtures/scenes/camera-pan-zoom.ts`: builds a minimal world with one non-static camera wired to **real** input bindings (`MouseAxis1dBinding` for wheel-zoom, `KeyboardAxis2dBinding` for arrow-key pan) - the same wiring a real game uses via `registerInputs`, not values poked directly into the actions. Renders a checkerboard grid of tinted squares (generated at runtime via an offscreen canvas, no static asset file) so pan/zoom is actually visible in a recording, not just a flat clear color.
- `e2e/specs/camera-pan-zoom.spec.ts` (Playwright, Chromium + SwiftShader for headless WebGL2): dispatches real `wheel`/keyboard events over several real-time-spaced frames and asserts **both** the camera's logic state (`zoom`/`position`) **and** what's actually rendered - `measureGreenSquareBounds()` scans the live canvas for the origin square's on-screen pixel bounds, and each test asserts the rendered change (size for zoom, position for pan) matches what the logic state predicts. This catches bugs a logic-only assertion can't: an earlier version of this scene was missing `createTransformEcsSystem`, so the camera's `position.local` updated correctly on pan but never propagated to `position.world` (what the renderer actually reads) - the camera never visibly moved, and only the rendered-pixel assertion caught it. Each test is broken into named `test.step()`s that print live as they run.
- New `.claude/skills/write-e2e-test` skill codifying this standard for future features: when a feature needs e2e coverage, why relative same-run pixel measurements (not absolute pixel/color values, not golden-image screenshot diffs) are the right tool, and how to wire one up.
- `handle.step()` advances the scene deterministically (drives `Time`/`EcsWorld` directly) instead of relying on `requestAnimationFrame`, so the suite is flake-free.
- New scripts: `npm run test:e2e`, `test:e2e:ui`, `check-types:e2e`, `dev:e2e`. `pretest:e2e`/`pretest:e2e:ui` print an immediate timestamp so a run is never silently "stuck" from the first moment it's invoked.
- **Wired into CI**: a new `test-e2e` job in `.github/workflows/ci.yml` runs the suite on every PR into `dev`, uploading the HTML/trace/video report as a workflow artifact on failure. Add `test-e2e` to the repo's required status checks (Settings → Branches) so a red e2e suite blocks merging.
- Locally, `trace` and `video` capture are `'on'` for every run (not just failures), so `npx playwright show-report e2e/playwright-report` always has a full replay/video available. CI stays `'retain-on-failure'`.
- `@playwright/test` is pinned to an exact version (`1.56.0`, not `^`) to match the browser binary revision available in this repo's dev/CI environments.
- **rendering:** `createRenderContext`/`RenderContext` gain a `preserveDrawingBuffer` option (default `false`, matching prior behavior), for consumers - like this suite's pixel measurements - that need to read the canvas back well after a frame has already been presented.
- `AGENTS.md` gets a new "Integration & E2E Testing" section documenting the layout, how to add a new scenario, the CI wiring, and gotchas future scenes will hit, including the visual-assertion pattern and pixel-reading pitfalls.

**Note on rendering-pixel assertions**: an earlier version of this PR included a test that read back the canvas's clear color as a "did the render pipeline actually run" smoke check. It was removed after diagnosis showed a deterministic, whole-canvas-wrong-color rendering result specific to this repo's CI environment's SwiftShader build at the time - reproduced identically via two independent readback methods (`gl.readPixels` and a `drawImage`/`getImageData` canvas readback), and unaffected by worker concurrency. Documented as a warning in `AGENTS.md` for any future pixel-based assertions.

A later round of this PR reintroduced pixel-based assertions (see above) and hit a second, unrelated CI-only failure: `dev` merged a breaking camera-projection change (#533, world units) while this PR was in flight, and GitHub Actions checks out the PR's *merge commit* for `pull_request`-triggered workflows by default, so CI was already exercising this scene against the new camera model even though local runs (against the pre-merge base) still passed. Diagnosed and fixed by merging `dev` into this branch and pinning the scene's camera to `verticalWorldUnits: canvas.height`, restoring the 1-world-unit-per-pixel assumption the scene's grid and spec math depend on.

Not included here (deliberate follow-up): additional scenes for input/physics/particles/animation beyond this first camera one.

## Test plan

- [x] `npm run check-types` - passes
- [x] `npm run check-types:e2e` - passes
- [x] `npm test` - all unit tests pass
- [x] `npm run test:e2e` - all 3 Playwright specs pass repeatedly, no flakes, locally and in CI (including the visual pixel-bounds assertions for zoom-in, zoom-out, and pan)
- [x] `npm run lint` - 0 errors (pre-existing warnings only, unrelated to this change)
- [x] `npm run cspell` - 0 errors
- [x] `npm run check-exports` - passes
- [x] `npm run build` - passes
- [x] Verified the harness's error-surfacing path end-to-end (navigating to an unknown scene renders the actual error on-page and via `pageerror`)
- [x] Verified the pan test's visual assertion actually catches a regression: temporarily removed `createTransformEcsSystem` from the scene and confirmed the assertion fails as expected, then restored it
- [x] All CI checks green on the current head commit, including `test-e2e`

---
_Generated by [Claude Code](https://claude.ai/code/session_01EwVxCxZwmeNz4jUrcCMkDV)_


src/rendering/render-context.ts
```diff
@@ -43,12 +43,14 @@ export class RenderContext {
...
    * @param canvas - The canvas element.
    * @param clearStrategy - The strategy for clearing the render context (default: CLEAR_STRATEGY.blank).
+   * @param preserveDrawingBuffer - Whether to retain the drawing buffer after presentation instead of letting the browser clear it, required for reading back the canvas's pixels (e.g. `toDataURL`, `drawImage`) after a frame has already been presented (default: false, since most consumers never read the canvas back and the retained buffer costs GPU memory bandwidth).
    */
   constructor(
     shaderCache: ShaderCache,
     imageCache: ImageCache,
     canvas: HTMLCanvasElement,
     clearStrategy: CLEAR_STRATEGY_KEYS = CLEAR_STRATEGY.blank,
+    preserveDrawingBuffer: boolean = false,
```
stormmuller, line 53: is false the most common value here? Is this the right default or just the most convenient one?

stormmuller: `false` is the right default, not just the convenient one:

- It matches the WebGL spec's own default for `preserveDrawingBuffer` when the attribute is omitted from context creation.
- It's the better default for a real-time engine: `true` disables some browser double-buffering optimizations and costs extra GPU memory bandwidth every frame, since the buffer has to be preserved instead of swapped/discarded. That cost is only worth paying if something actually reads the canvas back.
- Checked every existing consumer in `/src` and `/demo` (`grep -rn "getContext('webgl2'\|readPixels\|toDataURL\|preserveDrawingBuffer"`) - none of them read the canvas back today. The only place in the repo that needs it is the new `camera-pan-zoom` e2e scene, which opts in explicitly (`createRenderContext(canvas, { preserveDrawingBuffer: true })`).

So `false` keeps the common (render-only) path at its normal cost, and the rare (readback) path opts in explicitly where it's actually needed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EwVxCxZwmeNz4jUrcCMkDV)_

---

PR #535, opened 2026-07-26
test(e2e): add real-browser input suite for keyboard, mouse, and gamepad

stormmuller:
## Summary

Adds a real-browser Playwright suite covering the input system end-to-end, per `AGENTS.md`'s "Integration & E2E Testing" and the `write-e2e-test` skill: three new scenes/specs (`keyboard-input`, `mouse-input`, `gamepad-input`), each exercising every action type against real DOM events (a mocked `navigator.getGamepads` for the polled gamepad source, since Playwright has no gamepad emulation).

- **Action types**: `TriggerAction`, `HoldAction`, `Axis1dAction`, `Axis2dAction` — driven by real `page.keyboard`/`page.mouse` events and a fake polled `Gamepad`, asserted against both ECS state and actual rendered pixels (not logic state alone).
- **Action reset types**: `actionResetTypes.zero` vs `noReset`, including the documented pitfalls of misusing each — a held key with the default `zero` reset "twitches" (single impulse, then a compensating twitch back on release), and a gamepad axis with `zero` reset reads correctly for exactly one frame then gets permanently stuck at `0` because the source's own change-detection cache never notices the reset happened underneath it. Both are asserted as regression tests, not just described.
- **Input groups**: `InputManager.setActiveGroup` gating, proven by binding two actions in different groups to the *same* physical key/button/axis and confirming only the active group's action responds.

### Bug found and fixed along the way

Writing the mouse group-gating test surfaced a real bug: `MouseInputSource`'s trigger handlers called `binding.action.trigger()` directly instead of going through `InputManager.dispatchTriggerAction`, bypassing active-group gating entirely — unlike `KeyboardInputSource`, which already routes correctly, and unlike the documented contract in `actions.md`. Fixed in `mouse-input-source.ts`, with a new unit test file (`MouseInputSource` had none before) covering the regression, plus the mouse e2e test that would otherwise fail without the fix (verified locally by reverting the fix and re-running — it fails as expected).

## Test plan

- [x] `npm run check-types`
- [x] `npm test` (1067 tests, including the new `mouse-input-source.test.ts`)
- [x] `npm run lint` (0 errors)
- [x] `npm run cspell` (0 issues)
- [x] `npm run check-exports`
- [x] `npx eslint`/`tsc --noEmit --project e2e/tsconfig.json` on all new `/e2e` files
- [x] Full `/e2e` suite (16 tests across 4 spec files) run twice locally, all green
- [x] Confirmed the mouse group-gating test actually fails without the source fix (regression coverage verified, not assumed)
- [x] `documentation-site`: `npm run build` + `npm run typecheck` pass; manually loaded the `space-shooter` demo (uses `/input`) in a headless browser — renders cleanly, no console errors, keyboard input still drives the ship
- [x] `CHANGELOG.md` updated under `[Unreleased] > Fixed`
- [x] `AGENTS.md` updated with the gamepad-mocking pattern and the polling-cache step-timing gotcha for future e2e authors

---
_Generated by [Claude Code](https://claude.ai/code/session_01HG9K1WR1VXdsUGDCbzh64L)_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -477,33 +477,6 @@ If you add a pixel-reading assertion:
   See the `write-e2e-test` skill and `measureGreenSquareBounds()` in
   `camera-pan-zoom.ts` for the full pattern and rationale.
 
-**Testing a polled input source (`GamepadInputSource`) needs a fake
-device, and careful step accounting.** Playwright/CDP has no gamepad
-emulation and CI has no real controller, so `gamepad-input.ts` overrides
-`navigator.getGamepads` (via `Object.defineProperty`, before constructing
-`GamepadInputSource`) to return a fake `Gamepad` whose `axes` array a test
-hook mutates directly - this models Chrome's real behavior of updating the
-`Gamepad` object in place, which is what lets `GamepadInputSource.update()`
-see new values on its next poll. Unlike `KeyboardInputSource`/
-`MouseInputSource`, which only dispatch on a real DOM event,
-`GamepadInputSource` polls every frame and only re-dispatches when *its
-own* reading changed since the previous poll (see its class doc comment) -
-so an extra, unaccounted-for `step()` changes which frame a
-value-dependent assertion lands on. `camera-pan-zoom.spec.ts`'s
-`captureState` pattern (steps once *and* reads in the same helper) is safe
-for event-driven sources but wrong for a polled one: `gamepad-input.spec.ts`
-splits it into a `step()` and a separate no-step `readState()`, called 1:1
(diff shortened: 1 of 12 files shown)


AGENTS.md
```diff
@@ -477,6 +477,33 @@ If you add a pixel-reading assertion:
   See the `write-e2e-test` skill and `measureGreenSquareBounds()` in
   `camera-pan-zoom.ts` for the full pattern and rationale.
 
+**Testing a polled input source (`GamepadInputSource`) needs a fake
```
stormmuller: This is way too verbose for an agents file. Remove it.

PR #552, opened 2026-08-03
feat(math)!: migrate Vector2/Vector3 from classes to mutable plain objects

### LINKED ISSUE #537: Migrate Vector2/Vector3 from classes to plain objects + standalone functions
## Summary

`Vector2` (`src/math/vector2.ts`) and `Vector3` (`src/math/vector3.ts`) are currently classes with instance methods (`add`, `subtract`, `multiply`, `divide`, `normalize`, `rotate`, `dot`, `cross`, `clone`, `equals`, etc.) plus static getters for constants (`Vector2.zero`, `Vector2.up`, ...).

This issue proposes migrating both to **plain JS/TS objects** (e.g. `interface Vector2 { x: number; y: number }`) with all current instance methods converted into **standalone functions** (e.g. `add(a, b)`, `normalize(v)`, `dot(a, b)`), following the same "data is a plain object, behavior is a free function" philosophy already used for ECS components (see `AGENTS.md` § Component Pattern).

## Motivation

- **Consistency** with the rest of the engine's data/behavior split (components are plain interfaces + `add<Name>Component` factories, not classes).
- **Performance**: plain object literals avoid prototype chain / constructor overhead and are friendlier to V8 hidden-class optimization and GC when created in hot loops (physics, movement, particles).
- **Serialization/structural simplicity**: plain objects are trivially `JSON.stringify`-able, easy to pool, and easy to spread/copy (`{ ...v }`) without needing `clone()`.
- **Tree-shaking**: standalone functions allow bundlers to drop unused vector operations; a class ships every method whether used or not.

## Current scope

- `Vector2` is used across **51** non-test files in `/src`.
- `Vector3` is used across **2** non-test files in `/src`.
- Static constant getters (`.zero`, `.one`, `.up`, `.down`, `.left`, `.right`, `.forward`/`.backward` for `Vector3`) need standalone equivalents (e.g. exported constants or factory functions).
- Methods to convert: `set`, `add`, `subtract`, `multiply`, `multiplyComponents`, `divide`, `magnitude`, `magnitudeSquared`, `normalize`, `floorComponents`, `clone`, `toString`, `equals`, `toFloat32Array`, `distanceTo`, `rotate`, `dot`, `cross`, `perpendicular`, `negate`.
- Related helpers already living alongside the classes as standalone functions (`radians-to-vector.ts`, `vector-to-radians.ts`, `scale-relative-to-point.ts`, `smooth-damp-vector2.ts`) should be reviewed for consistency once the shape of `Vector2`/`Vector3` changes.

## Open question: return-new vs. mutate-in-place

The current class methods are **all immutable** (`add`, `subtract`, etc. return a *new* `Vector2`/`Vector3`), except `set()` which mutates in place. When converting to standalone functions we need to decide the convention. Weighing both:

### Option A — always return a new vector (pure functions)

```ts
const c = add(a, b); // a and b untouched
```

**Pros**
- Matches current behavior exactly → lowest-risk, most mechanical migration.
- No aliasing bugs: callers never have to worry about a function silently mutating an object they still hold a reference to elsewhere (e.g. a component's `position` field).
- Easier to reason about and test; composes naturally (`normalize(add(a, b))`).
- Safe to share/store vector objects (e.g. as component state) without defensive copying at read sites.

**Cons**
- Allocates a new object per operation. In hot paths (physics integration, per-entity-per-frame movement, particle updates) this creates significant GC pressure — potentially worse than the class version, since a class instance and a plain object allocate similarly, so this alone isn't a regression, but it doesn't fix the underlying perf concern that partly motivates the migration.
- Requires callers to reassign (`position = add(position, velocity)`), which is easy to forget and, unlike `Vector2.add`, gives no compiler hint that the original went untouched (mutation-by-omission bugs elsewhere already show up as "I called `.add()` and expected it to change the object").

### Option B — mutate in place, no return (or return the same object for chaining)

```ts
addInPlace(a, b); // a is now a + b
```

**Pros**
- Zero allocations for hot-path math — matches the perf-sensitive contexts vectors are heavily used in (physics, transform updates every frame).
- Familiar to engines like Three.js / Babylon.js, which default to in-place math methods precisely for this reason.

**Cons**
- Reintroduces the exact aliasing hazards classes already have today with `set()` — a shared reference (e.g. a `Vector2` stored directly on a component and also cached elsewhere) can be mutated unexpectedly.
- Breaks the "plain data object" mental model somewhat — functions with side effects on their arguments are more surprising for a value type than for a class instance with methods.
- Larger migration risk: every current call site assumes immutability; naively swapping to in-place semantics would silently change behavior rather than just changing syntax.

### Option C — hybrid: both, with a clear naming convention

Provide pure versions (`add`, `subtract`, ...) for general use and `*InPlace`/`addTo`-style mutating variants for identified hot paths (physics integration, per-frame transform updates), similar to how some math libraries (e.g. `gl-matrix`) expose both `vec2.add(out, a, b)`-style output-param functions and simpler pure wrappers.

**Pros**
- Lets call sites opt into zero-allocation math only where profiling shows it matters, without forcing every call site to deal with mutation semantics.
- Incremental: most of the 51 call sites can move to the pure functions with a near-mechanical `v.add(x)` → `add(v, x)` rewrite; only genuinely hot loops need the in-place variant.

**Cons**
- Two ways to do the same thing increases API surface and cognitive load; naming convention must be picked carefully and enforced (ESLint rule? code review?) to avoid inconsistent usage.
- Still need to decide the "third value" pattern (`out` parameter à la `gl-matrix`, e.g. `add(out, a, b)`) vs. `*InPlace(a, b)`, which have different call-site ergonomics and error-proneness (`gl-matrix`'s `out`-first convention is powerful but notoriously easy to misuse — e.g. `add(a, a, b)` aliasing `out` with an input).

**This issue is intentionally scoped to design + migrate; the specific convention (A, B, or C, and if C, the exact naming/signature convention) should be decided and documented in the design doc/PR before implementation starts**, since it affects call-site ergonomics across the whole engine (physics, rendering, transforms, input, animations all consume `Vector2`).

## Suggested approach

1. Design doc / RFC deciding the mutation convention (see above) and the final function signatures, naming, and file layout (likely `src/math/vector2/*.ts` + `src/math/vector2/index.ts`, mirroring how other math helpers are already standalone files in `src/math`).
2. Decide how constants (`Vector2.zero`, `.up`, etc.) are exposed as plain objects — likely frozen exported constants (`export const ZERO: Vector2 = Object.freeze({ x: 0, y: 0 })`) or factory functions if freezing is undesirable for perf/mutation reasons.
3. Implement `Vector2`/`Vector3` as plain interfaces + standalone functions, with full unit test parity with the existing `vector2.test.ts`/`vector3.test.ts`.
4. Migrate all 51 (`Vector2`) + 2 (`Vector3`) call sites in `/src`, plus tests, demos (`documentation-site/src/pages/demos`), and the `/demo` app.
5. Update `documentation-site/docs/docs` wherever `Vector2`/`Vector3` usage is documented.
6. Add a `CHANGELOG.md` entry under `[Unreleased]` → `#### Changed` (and `#### Removed` if the class export is dropped outright) since this is a breaking API change for consumers of the package.

## Breaking change

This is a breaking change to the public API (`Vector2`/`Vector3` currently ship as classes with `new Vector2(x, y)` construction and `instanceof` usable by consumers). Should be called out clearly in the changelog and likely warrants a major/minor version bump per semver once the engine is ≥ 1.0.0 (currently pre-1.0, per `CLAUDE.md`, so no upgrade guide is required yet, but the changelog entry is still required).

stormmuller:
## Summary

Closes #537.

Migrates `Vector2`/`Vector3` from classes to plain `{ x, y }`/`{ x, y, z }` objects, using the **mutable, in-place** convention (issue #537's "Option B") for the standalone math functions:

- `new Vector2(x, y)` / `new Vector3(x, y, z)` → `createVector2(x, y)` / `createVector3(x, y, z)`
- `Vector2.zero`/`.one`/`.up`/`.down`/`.left`/`.right` (and the `Vector3` equivalents plus `.forward`/`.backward`) → `vector2Zero()`/`vector2One()`/etc., factory functions that return a fresh object every call
- Arithmetic/transform methods (`add`, `subtract`, `multiply`, `multiplyComponents`, `divide`, `normalize`, `floorComponents`, `rotate`, `perpendicular`, `negate`, `set`) → standalone `vector2Add`/`vector2Rotate`/etc. functions that **mutate their first argument in place** and return it, instead of allocating a new vector — this is the perf motivation from the issue (avoids per-frame vector allocation in hot loops like physics integration and collision detection)
- Read-only operations (`magnitude`, `magnitudeSquared`, `dot`, `cross`, `distanceTo`, `equals`, `toString`, `toFloat32Array`) and `clone` → standalone functions that never mutate their arguments
- Prefixed with `vector2`/`vector3` (rather than flattened generic names like `add`) to avoid barrel-export collisions between the two types

### Mutation-safety pass

Since operations now mutate in place, every converted call site needed a check for whether the code relied on the input staying unchanged (as called out in the issue). Cloned with `vector2Clone`/`vector3Clone` wherever needed — live component fields (e.g. `position.world`, `rigidBody.velocity`) read without being reassigned, persistent warm-start/joint state reused across solver iterations or ticks, function parameters, and values reused more than once in the same scope. Left mutating in place where that's the intended fast path (e.g. integrating velocity into `position.world` every tick).

This surfaced and fixed a few latent aliasing bugs where a value was reused multiple times but a naive in-place conversion would have corrupted it (e.g. `sampleCubicBezier`'s reused Bezier control points, `applyPointImpulse`/`applyImpulse` being called with the same impulse for both sides of a constraint, an input `Axis2dAction`'s live value being scaled in place in a demo's movement system).

### Scope

- Core engine (`/src`): math, common, ecs, input, animations, particles, physics, rendering — all migrated, fully typechecked and tested.
- `/demo` app and `/e2e` fixtures: migrated.
- `documentation-site/src/pages/demos/**`: all demos migrated; verified with `tsc --noEmit`, a full `docusaurus build` (validates all doc links via `onBrokenLinks: 'throw'`), and manually checking all 17 demo pages in a headless browser (no console errors, correct rendering).
- `documentation-site/docs/docs/**`: updated all pages referencing the old `Vector2`/`Vector3` API (math guides, physics guides, particles, input, animations, lifecycle, rendering).
- `CHANGELOG.md`: added `Unreleased` entries under `Changed`/`Removed` (breaking change).

## Test plan

- [x] `npm run check-types` — 0 errors
- [x] `npm test` — 946 tests passed
- [x] `npm run lint` — 0 errors
- [x] `npm run cspell` — 0 issues
- [x] `npm run check-exports` — all subpaths green
- [x] `npm run check-types:e2e` — 0 errors
- [x] `npm run build` (repo root) — succeeds
- [x] `documentation-site`: `tsc --noEmit`, `npm run build` (full docusaurus build incl. typedoc), and manual browser check of all 17 demo pages — all render correctly with no console errors

---
_Generated by [Claude Code](https://claude.ai/code/session_01XCauuUDxyVDqYk7DA8PHHV)_


src/math/vector2.ts
```diff
@@ -1,265 +1,291 @@
...
-  public set(value: Vector2): this {
-    this.x = value.x;
-    this.y = value.y;
+/**
+ * Creates a new Vector2.
+ * @param x - The x-coordinate component (default: 0).
+ * @param y - The y-coordinate component (default: 0).
+ * @returns The created vector.
+ */
+export const createVector2 = (x: number = 0, y: number = 0): Vector2 => ({
```
stormmuller: This create function offers not value. Remove it.

src/math/vector2.ts
```diff
@@ -1,265 +1,291 @@
...
+  y,
+});
 
-    return this;
-  }
+/**
+ * Creates a zero vector (0, 0).
+ * @returns A new Vector2.
+ */
+export const vector2Zero = (): Vector2 => createVector2(0, 0);
```
stormmuller: I don't like the DX of all of these `vector2XYZ` functions. 

Let's create a static `Vec2` class that has these constructor functions as static members.

`Vec2.Zero`

src/math/vector2.ts
```diff
@@ -1,265 +1,291 @@
...
+ * Normalizes `target` to unit length in the same direction, mutating it in place.
+ * @param target - The vector to mutate.
+ * @returns `target`, for chaining.
+ */
+export const vector2Normalize = (target: Vector2): Vector2 => {
+  const length = vector2Magnitude(target);
 
-    return this.divide(length);
+  if (length === 0) {
+    return target;
```
stormmuller: A nomalized vector with a length of zero is undefined. Throw an error.

src/math/vector2.ts
stormmuller: Removed. `createVector2`/`createVector3` are gone entirely, replaced by `Vec2.create`/`Vec3.create` as part of the static-class restructuring below.

---

src/math/vector2.ts
stormmuller: Done — `vector2X`/`vector3X` are gone, replaced by `Vec2`/`Vec3` classes holding every operation as a `public static` method. `Vector2`/`Vector3` stay as the plain `{ x, y }`/`{ x, y, z }` data interfaces (kept as separate names from `Vec2`/`Vec3` to avoid TS declaration-merging between a class and an interface of the same name).

I asked a clarifying question on a couple of specifics before doing this pass (mainly the constant-factory shape) and didn't hear back, so I went with:
- Constants (`zero`, `one`, `up`, `down`, `left`, `right`, plus `forward`/`backward` on `Vec3`) as lowercase `static get` accessors rather than `Zero`/`PascalCase`, to match this repo's camelCase-for-public-members convention — `Vec2.zero`, `Vec2.up`, etc.
- Each constant getter returns a **fresh** vector on every access (not a shared/frozen singleton), since the whole point of this migration is that vectors are mutable — a shared `Vec2.zero` instance would get silently corrupted the first time any caller mutated it in place.
- `Vec2.create`/`Vec3.create` replace the removed `createVector2`/`createVector3` (see the other thread).

Happy to adjust the naming/shape if this isn't quite what you had in mind.

---

src/math/vector2.ts
stormmuller: Done — `Vec2.normalize`/`Vec3.normalize` now throw (`Unable to normalize a zero-length Vector2/Vector3.`) instead of silently returning the zero vector.

---

src/math/vector2.ts
stormmuller: No I don't want a create function at all. The interface is easy to initialize with a literal object.

stormmuller: Understood — `Vec2.create`/`Vec3.create` are gone. Every call site now constructs `Vector2`/`Vector3` with a plain object literal (`{ x, y }`/`{ x, y, z }`) instead, including `Vec2`/`Vec3`'s own internal constant getters (`zero`, `one`, `up`, etc.).

---

PR #587, opened 2026-08-06
fix(common): apply parent rotation and scale to child position offsets

### LINKED ISSUE #581: fix(common): parented position offsets ignore the parent's rotation and scale
## Summary

`composeWithParent` in `src/common/systems/transform-system.ts` inherits rotation and scale correctly, but composes a child's world **position** by plain addition. The child's local offset is never rotated or scaled by the parent's world transform, so a child of a rotating parent spins in place instead of orbiting it.

Raised by @stormmuller while reviewing #580.

## Current behaviour

`src/common/systems/transform-system.ts`:

```
world.rotation = parent.world.rotation + local.rotation      // correct
world.scale    = parent.world.scale    * local.scale         // correct
world.position = parent.world.position + local.position      // missing two terms
```

A correct 2D TRS composition transforms the child's local offset by the parent before adding it:

```
world.position = parent.world.position
               + rotate(local.position * parent.world.scale, parent.world.rotation)
```

## Reproduction

Parent at the origin with `rotation.local = Math.PI / 2`, child parented to it with `position.local = { x: 10, y: 0 }`:

| | Expected | Actual |
| --- | --- | --- |
| child `position.world` | `(0, 10)` | `(10, 0)` |
| child `rotation.world` | `π/2` | `π/2` ✓ |

The child's rotation is right, so the sprite visibly spins — it just doesn't orbit. The scale case fails the same way: parent `scale.local = (2, 2)` and child `position.local = (10, 0)` gives `(10, 0)` where it should give `(20, 0)`.

Symptom in a real scene: a turret parented to a rotating tank rotates correctly but stays glued at an un-rotated offset, so it detaches from its mount as the tank turns.

## Why it wasn't caught

`transform-system.test.ts` has seven tests and none of them attach a `RotationEcsComponent` or `ScaleEcsComponent` to a parent. Nothing currently asserts this behaviour in either direction.

## Related dead code

`parent-position-system.ts`, `parent-rotation-system.ts`, and `parent-scale-system.ts` (all in `src/common/systems/`) carry the same additive position composition. They are **not** exported from `src/common/systems/index.ts` and are referenced only by their own test files — they appear to be superseded by `createTransformEcsSystem`. Worth deleting alongside the fix rather than fixing in three places, unless they're being kept deliberately.

## Suggested scope

- Fix `composeWithParent` to apply the parent's world rotation and scale to the child's local offset.
- Add tests covering: rotated parent, scaled parent, rotated + scaled parent, and a three-deep chain.
- Decide whether to delete the three superseded `parent-*-system.ts` files and their tests.

Note this is a **behaviour change** for any existing content that parents a positioned entity to a rotated or scaled one, and that has been authored against the current (incorrect) composition. Given the engine is pre-1.0, correcting it seems right, but it likely wants a changelog entry flagging the visual change.

stormmuller:
## Summary

`composeWithParent` in `src/common/systems/transform-system.ts` inherited a parent's world rotation and scale correctly, but composed a child's world **position** by plain addition of the parent's world position. The child's local offset was never rotated or scaled by the parent's world transform, so a child of a rotating parent spun in place instead of orbiting it (and a child of a scaled parent kept an un-scaled offset).

- Split `composeWithParent` into `composePositionWithParent`, `composeRotationWithParent`, and `composeScaleWithParent` (also fixes a SonarJS cognitive-complexity lint failure the fix would otherwise trigger).
- `composePositionWithParent` now rotates and scales the child's local offset by the parent's world rotation/scale before adding it to the parent's world position:
  `world.position = parent.world.position + rotate(local.position * parent.world.scale, parent.world.rotation)`
- Deleted `parent-position-system.ts`, `parent-rotation-system.ts`, and `parent-scale-system.ts` (and their tests) - these carried the same additive-position bug, were superseded by `createTransformEcsSystem`, and were not exported from `src/common/systems/index.ts` or referenced anywhere outside their own tests.
- Added tests covering a rotated parent, a scaled parent, a rotated + scaled parent, and a three-deep parent chain.

No demo under `/documentation-site/src/pages/demos` uses `addParentComponent`/`ParentEcsComponent`, so none are affected by this behavior change.

**Behaviour change**: any existing content that parents a positioned entity to a rotated or scaled one, authored against the previous (incorrect) additive composition, will now render differently (correctly). Flagged in the changelog per the issue's suggestion.

## Related issue(s)

Closes #581

## Verification checklist

- [x] `npm test` passes
- [x] Any new/changed public API is exported from the module's `index.ts` (and `/src/index.ts` / `package.json` `exports` if it's a new module) — n/a, no public API changed
- [x] Documentation under `/documentation-site/docs/docs` is updated if this change affects documented behavior — n/a, no doc describes this composition behavior
- [x] If this change touches a module with a demo under `/documentation-site/src/pages/demos`, the demo has been updated and verified — n/a, no demo uses entity parenting

## Changelog

- [x] A bullet has been added under `## [Unreleased]` in `CHANGELOG.md`

---
_Generated by [Claude Code](https://claude.ai/code/session_01M42FeAcwfsHYR3btuySY5U)_


stormmuller:
CodeQL analysis passed. However, the `CI / Build and Test` workflow (`.github/workflows/ci.yml`) and the `Changelog` workflow don't appear to have triggered at all for this PR - no run record exists for this branch, and neither workflow has run repo-wide since ~17:46 UTC today (Aug 6), several hours before this PR was opened. The workflow definitions are active/unchanged, so this looks like an infra-side issue (e.g. Actions runner/queue backlog or an account-level quota) rather than something in this PR's diff. Flagging since it's blocking the required status checks for auto-merge - may need a look from someone with access to the org's Actions settings/usage.

---

PR #590, opened 2026-08-07
fix(rendering): make SpriteEcsComponent.pivot Y-up to match the rest of the engine

### LINKED ISSUE #585: fix(rendering): SpriteEcsComponent.pivot is Y-down while the rest of the engine is Y-up
## Summary

`SpriteEcsComponent.pivot` uses a Y-**down** convention — `(0, 0)` is the sprite's **top**-left — while world space, entity positions, and rotation are all Y-**up**. Pivot is the only part of the public surface that inverts, and it does so because an internal projection detail leaks through.

Raised by @stormmuller while reviewing #580.

## Confirming the convention

The JSDoc on `SpriteEcsComponent.pivot` states `(0, 0)` is the top-left corner, and tracing the pipeline confirms the code matches that:

1. **`createProjectionMatrix`** scales Y by `-2 / height` — the projection flips Y.
2. **`bindSpriteInstanceData`** pre-negates the entity's position: `buffer[POSITION_Y_OFFSET] = -position.world.y`. Those two negations cancel, so a larger `position.world.y` correctly draws **higher** on screen. World space is Y-up. ✅
3. **`sprite.vert.glsl`** computes the quad's local offset *in the shader*, so it never gets that compensating negation:
   ```glsl
   vec2 normalizedPivot = (a_instancePivot - 0.5) * 2.0;
   vec2 pivoted = a_position - normalizedPivot;
   vec2 scaled  = pivoted * a_instanceSize * a_instanceScale * 0.5;
   vec2 world   = rotated + a_instancePos;
   ```
   Local `+y` therefore passes through the projection's flip unopposed and moves **down** the screen.

With `pivot = (0, 0)`: `normalizedPivot = (-1, -1)`, so `pivoted = a_position + 1` spans `[0, 2]`, and `scaled` spans `[0, size]`. All local offsets are ≥ 0, and since local `+y` is down, the sprite hangs **below** the entity position — i.e. the entity sits at the sprite's **top** edge. Top-left, as documented.

## Why it's a bug

Every other Y-facing convention in the engine is Y-up, and each is deliberately corrected for the projection flip:

| | Convention | Compensated? |
| --- | --- | --- |
| `position.world.y` | Y-up | ✅ negated in `bindSpriteInstanceData` |
| `RotationEcsComponent` | Y-up | ✅ negated, with a comment explaining it matches position |
| `screenToWorldSpace` | Y-up | ✅ flips Y explicitly |
| **`SpriteEcsComponent.pivot`** | **Y-down** | ❌ **not compensated** |

So pivot is the one place a rendering-internal detail surfaces in the public API. A caller reasoning "my world is Y-up, so pivot `(0,0)` is bottom-left" gets a sprite offset vertically by its full height — and the error is invisible at the default centered pivot, which is why it has survived.

## Suggested fix

Negate the pivot's Y contribution so `(0, 0)` means bottom-left, then update the JSDoc. Either flip on upload:

```ts
instanceDataBufferArray[offset + PIVOT_Y_OFFSET] = 1 - sprite.pivot.y;
```

or handle it in `sprite.vert.glsl` alongside the existing `normalizedPivot` conversion. The shader is arguably the better home, since that's where the flip originates.

**This is a breaking behaviour change** for any content using a non-centered pivot — the default `(0.5, 0.5)` is unaffected, which limits the blast radius. Pre-1.0, correcting it seems right, but it wants a changelog entry calling out the visual change.

## Worth checking in the same pass

- **Nine-slice region offsets.** `render-system.ts` adds `region.offset` directly to `entityPosition.world` (Y-**up** space) while the pivot inside the shader is Y-**down**. Those two spaces disagree, so `computeNineSliceRegions` should be re-derived against whichever convention wins. Nine-slice always renders its regions with a centered pivot, which may be masking this.
- **Test coverage.** Given [the pivot bug fixed in 0.24.1](https://github.com/Forge-Game-Engine/Forge/blob/dev/CHANGELOG.md) (pivot applying only half its offset), non-centered pivots look under-tested. Worth adding cases that pin all four corners.

## Relationship to #580

`design/ui-system.md` §5.2 documents the mismatch as a gotcha and specifies a `sprite.pivot.y = 1 - rectTransform.pivot.y` bridge in the UI layout system. If this is fixed, that bridge becomes unnecessary and should be deleted rather than kept as a compensating error.

stormmuller:
## Summary

`SpriteEcsComponent.pivot` used a Y-**down** convention (`(0, 0)` was the sprite's **top**-left) while world position, rotation, and every other Y-facing value in the engine are Y-**up**. The mismatch was an internal projection detail (the vertex shader flips Y before the projection matrix flips it back for `position`/`rotation`) leaking through pivot's public API - invisible at the default centered pivot, which is why it survived.

Traced the full pipeline per the issue and confirmed the analysis: `bindSpriteInstanceData` pre-negates `position.world.y` and `rotation.world` to compensate for the projection's Y flip, but `sprite.vert.glsl` computes the pivot-based local offset *after* that compensation, so it never gets flipped - `pivot.y = 0` landed at the sprite's top instead of its bottom.

## Changes

- **`sprite.vert.glsl`**: negate `a_instancePivot.y` alongside the existing `[0,1] -> [-1,1]` doubling, so pivot's Y now gets the same flip `position`/`rotation` already get. `(0, 0)` is now bottom-left, `(1, 1)` is top-right.
- **`computeNineSliceRegions`** (`compute-nine-slice-regions.ts`): updated the region-offset math (`(1 - pivot.y) * height - ...` instead of `pivot.y * height - ...`) so nine-slice region placement agrees with the corrected shader convention. The centered pivot `(0.5, 0.5)` produces identical output before and after (`1 - 0.5 === 0.5`), so this is a no-op for every sprite in the repo (no demo or test uses a non-centered pivot).
- **`sprite-component.ts`**: updated `pivot`'s JSDoc to describe the new Y-up convention.
- **`compute-nine-slice-regions.test.ts`**: replaced the single top-left-pivot test with a parameterized case pinning all four corners (bottom-left, bottom-right, top-left, top-right), per the issue's test-coverage callout.
- **`CHANGELOG.md`**: added a `Fixed` entry under `[Unreleased]`, flagged as a breaking change for non-centered pivots.

## Related issue(s)

Closes #585

## Out of scope

The issue's design doc reference (`design/ui-system.md`'s pivot bridge, from #580) isn't in `dev` yet, so there's nothing to remove there in this PR.

## Verification checklist

- [x] `npm test` passes (1036 tests)
- [x] `npm run lint` passes with 0 errors (pre-existing, unrelated TODO warnings only)
- [x] No public API surface changed beyond `pivot`'s documented semantics - no export changes needed
- [x] Documentation under `/documentation-site/docs/docs` - no page documents the pivot convention, so nothing to update
- [x] No demo uses a non-centered pivot (verified via `grep`), so no demo code changes; still built `/dist`, ran `documentation-site`'s `typecheck`/`build`, and visually checked the `nine-slice` and `brick-breaker` demos in a browser to confirm no regression at the default centered pivot

## Changelog

- [x] Added a bullet under `## [Unreleased]` → `#### Fixed` in `CHANGELOG.md`

---
_Generated by [Claude Code](https://claude.ai/code/session_01NMZekQ8g3g2CnrBpBuE5Y7)_


stormmuller:
In the ECS demo. The position looks correct, but the rotation looks inverted?

stormmuller:
Looked into this — this PR doesn't touch rotation at all, and for a centered pivot (which the ECS demo's `star` sprite uses, via `createImageSprite`'s default `pivot: { x: 0.5, y: 0.5 }`) the shader's `normalizedPivot` is `(0, 0)` both before and after this change (`(0.5 - 0.5) * 2 = 0`, negated or not), so the rendered output for that demo should be byte-for-byte identical to `dev`.

To double-check rather than just reason about it, I tracked the star's actual on-screen rotation with a frame-diff (rotate-and-correlate) measurement against two captured frames ~520ms apart on this branch's build: the star rotates ~30.5° **counter-clockwise** on screen in that window, matching `rotation.world`'s expected `+29.9°` change (`Δt · 1 rad/s`, since the demo sets `rotation.world = time.timeInSeconds` directly) and the engine's documented positive-rotation-is-CCW-in-Y-up-space convention (same convention `bindSpriteInstanceData`'s rotation negation targets, and the same one physics/`Vector2.rotate` use elsewhere).

So the rotation direction looks correct to me, and unrelated to this PR either way — happy to be pointed at a specific frame/timestamp if it still looks inverted to you, or if you meant something else by "rotation" (e.g. the orbit path direction, which is independent of the sprite's own spin and also unchanged by this diff).

---

PR #591, opened 2026-08-07
feat(text): add MSDF font atlas text rendering module

### LINKED ISSUE #584: feat(text): text rendering — MSDF font atlases, shaping, and a TextEcsComponent
## Summary

Forge cannot render text. `grep -ri "font\|fillText" src` returns only terrain-mesh noise — there is no glyph, no font asset type, no loader, no shaper, no text component.

Raised by @stormmuller while reviewing #580, on the grounds that text (like clipping, #583) is generally useful outside a UI module. **Explicitly out of scope for the UI work**; filed here so it can be built and released on its own merits.

Proposed home: a new `/src/text` module exporting `@forge-game-engine/forge/text`.

## Why it isn't a UI feature

Damage numbers, dialogue, floating names, debug overlays, and world signage all want text with no UI involved. Burying it under `/src/ui` would force those consumers to import a UI module they never use.

## Proposed approach: MSDF

| | Batches | Crisp when scaled | Toolchain | Effects |
| --- | --- | --- | --- | --- |
| Canvas2D → texture per string | ✗ one draw per string | ✗ | none | ✗ |
| Bitmap atlas (BMFont) | ✓ | ✗ blurs above authored size | atlas generator | ✗ |
| **MSDF** | **✓** | **✓ any scale** | atlas generator | **✓ outline/glow/shadow nearly free** |
| Vector tessellation | ✗ | ✓ | none | partial |

Canvas2D re-uploads a texture on every text change — fatal for something as ordinary as a score counter — and breaks batching entirely. MSDF's only real cost over a bitmap atlas is a ~10-line fragment shader (median of three channels, then `smoothstep`), and it buys arbitrary scaling plus the outline and glow effects game text actually wants.

## Suggested scope

- **`FontAtlas` type + `loadFontAtlas`** in `/src/asset-loading` — [`msdf-atlas-gen`](https://github.com/Chlumsky/msdf-atlas-gen) JSON + PNG. Glyph UVs, advances, **kerning pairs**, `lineHeight`, `baseline`, `distanceRange`. Kerning pairs must be present in the metrics JSON or text looks subtly wrong in a way that's hard to attribute later.
- **MSDF fragment shader** + a `createMsdfTextRenderable` helper, with screen-space-derivative antialiasing.
- **`TextEcsComponent`** (text, font, size, color, alignment, wrap width, line spacing) and **`TextMeshEcsComponent`** (the shaped glyph quads), plus `createTextShapingEcsSystem` — dirty-tracked, since re-shaping a paragraph every frame is genuinely expensive.
- **A default MSDF atlas shipped with the engine**, permissively licensed, so `createLabel(world, 'Hello')` works with zero toolchain setup. Document the generation recipe for custom fonts.
- Text effects (outline, drop shadow, glow) fall out as extra shader parameters and are cheap to add once the above exists.

## A deliberate Canvas2D escape hatch

MSDF is the shipping path, but the atlas generation step is a real barrier at exactly the wrong moment — someone trying their own font for the first time, or throwing a debug readout on screen mid-session. So **keep a Canvas2D-backed path as a documented prototyping escape hatch**: render a string to a texture with `fillText`, wrap it in an ordinary sprite.

Its limits should be stated plainly wherever it's documented, so nobody reaches for it in a shipping game by accident:

- One draw call per string — no batching.
- Re-uploads the texture on every text change, so it's unusable for anything that updates per frame (score, timer, FPS counter).
- Blurs when scaled, since it's a raster at one size.
- No outline/glow/shadow parameters.

The two paths should share the same `TextEcsComponent` surface where practical, so moving a prototype onto a real atlas is a font swap rather than a rewrite.

## Frame ordering constraints for `createTextShapingEcsSystem`

Where the shaping system sits in the tick is load-bearing, and easy to get subtly wrong:

- **Before `createRenderEcsSystem`.** The render pass consumes the glyph quads, so they must already exist. Register shaping with a lower `SystemRegistrationOrder` than the render system.
- **After any system that can change the wrap width.** For text laid out inside a UI container, the wrap width comes from the element's *resolved* rect, which only exists after the UI layout pass and `createTransformEcsSystem` have run. Shaping earlier means wrapping against last frame's width — invisible until something resizes, then one frame of visibly wrong line breaks.
- **No layout dependency for standalone text.** Damage numbers, floating names, and debug overlays have no container constraining them, so their shaping depends only on the string, font, and size. Those can shape anywhere before render.

The practical consequence: the ordering requirement is *conditional* on whether text is inside a laid-out container. Simplest correct default is to register shaping late — after layout and transform, before render — which is correct for both cases. `design/ui-system.md` §5.4 shows this placement in the full pipeline diagram.

**Dirty tracking is the other half of this.** Re-shape only when `text`, `font`, `fontSize`, or the resolved wrap width actually changed. Every other system in the UI design recomputes unconditionally per frame (see #580's DL-12); text shaping is the deliberate exception, because a paragraph re-shaped every frame is real cost rather than theoretical.

## The one renderer change this needs

One text element is many glyph quads. `render-system.ts` already expands **one** `SpriteEcsComponent` into **up to nine** render commands for nine-slice (`computeNineSliceRegions`), which is the same operation with fewer quads.

The suggestion in #580's DL-05 is to generalise that into a single `SubQuad[]` expansion path that both nine-slice and text produce, rather than giving text its own render system — a separate text pass could not interleave correctly with panels in one sorted command buffer, so a label would draw either always above or always below its own background.

**This is the highest-risk part of the work.** It touches a hot path with existing tests and the eight-plus demos that depend on nine-slice. Recommend landing it as a pure refactor with zero behaviour change first, verified against the nine-slice demo and the e2e suite, and only then adding text on top. Fallback if it proves thorny: a second expansion branch for text in the same function — uglier, but isolated.

## Relationship to the UI design (#580)

`design/ui-system.md` DL-04 and DL-05 hold the decision records that led here; that document now defers to this issue rather than owning the work. §5.4's pipeline diagram shows where the shaping system lands relative to layout, transform, and render.

**Dependency, stated plainly:** the UI module is not useful without text. Buttons need labels, HUDs need numbers. UI layout, anchoring, hit-testing, and interaction can all be built and unit-tested before this lands, but the UI cannot ship a convincing demo — or arguably a convincing release — until it does. This is the single largest dependency on the UI plan's critical path.

stormmuller:
## Summary

Adds a new `@forge-game-engine/forge/text` module implementing the MSDF (multi-channel signed distance field) text rendering path proposed in #584:

- **`FontAtlas` type + `loadFontAtlas`** (`/asset-loading`) - parses an [`msdf-atlas-gen`](https://github.com/Chlumsky/msdf-atlas-gen) JSON metrics file and loads its atlas PNG (via the existing `ImageCache`) into glyph advances, kerning pairs, and line metrics, all normalized to em units. Kept data-only (no WebGL) so `/asset-loading` keeps its existing zero-dependency-on-rendering direction.
- **MSDF fragment shader + `createMsdfTextRenderable`** (`/rendering`) - reuses the existing `sprite.vert` shader and its instance data layout (`spriteInstanceDataSegment`) completely unchanged, paired with a new `msdf.frag` shader (median-of-three-channels reconstruction + `fwidth`-based screen-space-derivative antialiasing, the standard technique from `msdf-atlas-gen`'s own docs). One renderable is built per `FontAtlas` and shared across every entity using that font, exactly like a sprite's `Renderable`.
- **`TextEcsComponent`/`addTextComponent`, `TextMeshEcsComponent`, and `createTextShapingEcsSystem`** (`/text`) - `shapeText` (pure, fully unit-tested) word-wraps, applies kerning, and lays out lines by alignment/line-spacing/pivot into glyph quads. The shaping system is dirty-tracked: it only re-shapes an entity when `text`, `font`, `fontSize`, `wrapWidth`, `lineSpacing`, `alignment`, or `pivot` actually changed, not every frame.
- **`createRenderEcsSystem` text integration** - per the issue's own risk analysis, this skips the larger "generalize nine-slice into a shared `SubQuad[]` expansion" refactor in favor of the issue's suggested fallback: an isolated `pushTextRenderCommands`/`buildTextCameraCommands` pair that turns each glyph into a `SpriteEcsComponent`-shaped render command. This reuses the sprite instancing/batching/depth-sort pipeline completely unchanged (zero modifications to the existing sprite/nine-slice code path) while still letting text interleave correctly with sprites in one sorted command buffer.

### Not included (follow-up work)

Flagged explicitly in the new docs' "Current limitations" section:

- **No font shipped with the engine** - generating a real MSDF atlas requires the `msdf-atlas-gen` toolchain, which isn't available in this environment. Every project currently supplies its own atlas.
- **No Canvas2D prototyping escape hatch** - the issue proposes a documented `fillText`-to-texture fallback for quick prototyping; not built here.
- **No text effects** (outline/glow/shadow) - these fall out as extra MSDF shader parameters later.
- **No demo** - without a real generated font atlas to point one at, a `/documentation-site/src/pages/demos/text` demo would have nothing genuine to render. Everything else in the module (shaping math, dirty tracking, shader, render-system integration) has full unit-test coverage instead.

## Related issue(s)

Closes #584

## Verification checklist

- [x] `npm test` passes (1084 tests, 161 files)
- [x] `npm run check-exports` passes (`@forge-game-engine/forge/text` resolves 🟢 ESM/bundler like every other subpath)
- [x] New public API is exported from each module's `index.ts`, `/src/index.ts`, and `package.json` `exports` (new `./text` entry)
- [x] Documentation added under `/documentation-site/docs/docs/text` (overview + an MSDF Text guide covering atlas generation, wrapping/alignment, registration order, and dirty tracking), plus updates to the `asset-loading` and `rendering` section docs to cross-link it
- [x] No existing demo touches `TextEcsComponent`/the new msdf shader, so none needed updating - but since this PR does touch `/rendering` (render-system.ts, the shader cache, shader index), I still ran `npm run build`, `documentation-site`'s `npm run typecheck`/`npm run build`, and loaded the Nine-Slice demo (the closest existing exerciser of the sprite render-command path this PR adds a sibling branch next to) plus the new `/docs/text` and `/docs/text/msdf-text` pages in a real browser (Playwright/Chromium) to confirm no regressions and no MDX/build errors

## Changelog

- [x] Added a bullet under `## [Unreleased]` → `#### Added` in `CHANGELOG.md`

---
_Generated by [Claude Code](https://claude.ai/code/session_0113gCpCEDFQad8Y9dJuahSq)_


stormmuller:
Added an interactive [Text demo](documentation-site/src/pages/demos/text) per the `add-demo` skill's own criteria (a new top-level `/src` module unambiguously needs one) — this was missing from the original PR.

Rather than skip it for lack of a font asset, built the actual [`msdf-atlas-gen`](https://github.com/Chlumsky/msdf-atlas-gen) C++ toolchain from source and generated a real MSDF atlas from Liberation Sans (SIL OFL 1.1, bundled under `documentation-site/static/fonts/liberation-sans/` with its license and a regeneration recipe). This exercises `loadFontAtlas` against genuine `msdf-atlas-gen` JSON output for the first time (previously only unit-tested against hand-built fixtures), which is good added confidence in the parser.

The demo shows word-wrap/kerning/alignment side by side (left/center/right, same paragraph) plus a live per-frame-updated counter label demonstrating `createTextShapingEcsSystem`'s dirty tracking. Verified in a real browser — screenshots confirmed crisp MSDF glyph rendering and the counter ticking frame to frame. `documentation-site`'s `typecheck` and `build` both pass.


---

PR #595, opened 2026-08-08
feat(text): add MSDF font atlas generation, format, and loading (phase 1)

stormmuller:
## Summary

Implements **Phase 1** of `design/msdf-text-rendering.md`: the offline atlas pipeline, on-disk format, and runtime loader. No rendering yet (that's Phase 2) — this is purely the data pipeline, fully testable without a GPU, per the design's own Phase 1 scope and Definition of Done.

- `npm run generate-font-atlas -- --font <path> --out <path> [--charset ascii|<string>|<file>] [--size 42] [--distance-range 4] [--texture-width 512] [--texture-height 512]` (`scripts/generate-font-atlas.mjs`) generates an MSDF atlas PNG plus a normalized, versioned metrics JSON from a `.ttf`/`.otf` font.
- `@forge-game-engine/forge/text` (new module, `/src/text`):
  - `FontAtlasData` / `FontAtlasFileData` — the runtime (`Map`-based) and on-disk (JSON-safe) shapes of a font atlas's metrics (`GlyphMetrics`, kerning, line metrics), versioned via `formatVersion` (DL-03).
  - `validateFontAtlasFileData` — throws a descriptive error on malformed/incompatible-version atlas JSON rather than failing deep inside a future shaping system (§5.13/§5.12).
  - `FontAtlasCache` (implements the existing `AssetCache<T>` contract) — loads an atlas's JSON + image (via the existing `ImageCache`) into a `FontAtlas`.
- Verified end-to-end against a real system font (DejaVu Sans): generated an atlas, then loaded and validated it through the actual `FontAtlasCache`/`validateFontAtlasFileData`/`toFontAtlasData` runtime path.

### Deviation from the design doc: generator tool

DL-02 proposed wrapping `msdf-atlas-gen` (native binary via its npm wrapper). That npm package transitively pulls in `google-font-metadata` → `puppeteer`, whose postinstall step downloads a full Chrome browser — a heavy, unrelated, and fragile dependency for a devDependency-only build tool (it failed outright in this sandbox and would be a flaky `npm ci` step in CI). Instead this wraps [`msdf-bmfont-xml`](https://github.com/soimy/msdf-bmfont-xml), which bundles prebuilt `msdfgen` binaries directly (same underlying reference implementation DL-02 calls for), has no network-dependent postinstall, and is called as a plain library function rather than shelled out to. `npm audit --omit=dev` shows 0 vulnerabilities from this addition.

## Related issue(s)

N/A

## Verification checklist

- [x] `npm test` passes (1062 tests, including 26 new)
- [x] New public API is exported from `src/text/index.ts`, `/src/index.ts`, and `package.json` `exports` (`./text`)
- [ ] Documentation under `/documentation-site/docs/docs` — intentionally not added; the design doc defers the conceptual guide to Phase 5, once there's a renderable feature to document
- [ ] No demo under `/documentation-site/src/pages/demos` references this module yet (n/a for this phase)

## Changelog

- [x] A bullet has been added under `## [Unreleased]` → `#### Added` in `CHANGELOG.md`

---
_Generated by [Claude Code](https://claude.ai/code/session_01SAHCUFuDsbsYiYGpt25AhC)_

### DIFF
--- a/documentation-site/docs/docs/text/generating-a-font-atlas.md
+++ b/documentation-site/docs/docs/text/generating-a-font-atlas.md
@@ -16,15 +16,19 @@ something your game runs at runtime.
 The command wraps [`msdf-bmfont-xml`](https://www.npmjs.com/package/msdf-bmfont-xml),
 an **optional peer dependency** of `@forge-game-engine/forge`. A plain
 `npm install @forge-game-engine/forge` does not install it, since most games
-don't need the atlas generator itself, only its output. Install it once,
-as a dev dependency of your own project:
+don't need the atlas generator itself, only its output. Install it as a dev
+dependency of your project:
 
 ```bash
 npm install --save-dev msdf-bmfont-xml
 ```
 
-Running the command without it installed fails fast with a message telling
-you to install it, rather than a raw module-not-found error.
+Or, since most games never need it at runtime and only run it occasionally,
+install it globally instead:
+
+```bash
+npm install -g msdf-bmfont-xml
+```
 
 ## Generate an atlas
 
@@ -32,20 +36,19 @@ you to install it, rather than a raw module-not-found error.
 npx forge-generate-font-atlas --font my-font.ttf --charset ascii --out assets/fonts/my-font
 ```
 
-This writes `assets/fonts/my-font.png` and `assets/fonts/my-font.json`.
-Commit both to your project; [`FontAtlasCache`](./loading-a-font-atlas.md)
-loads them together at runtime.
+This writes `assets/fonts/my-font.png` and `assets/fonts/my-font.json`, which
+[`FontAtlasCache`](./loading-a-font-atlas.md) loads together at runtime.
 
-| Flag                             | Default   | Meaning                                                                                                 |
-| --------------------------------- | --------- | --------------------------------------------------------------------------------------------------------- |
-| `--font <path>`                   | required  | The input `.ttf`/`.otf` file.                                                                              |
-| `--out <path>`                    | required  | Output path with no extension; writes `<out>.png` and `<out>.json`.                                        |
-| `--charset <name\|string\|path>`  | `ascii`   | `ascii` for printable ASCII (32-126), a literal string of characters, or a path to a text file to read them from. |
-| `--size <n>`                      | `42`      | The font size, in pixels, the distance field is authored at.                                               |
-| `--distance-range <n>`            | `4`       | The distance field's encoded range, in pixels.                                                             |
-| `--texture-width <n>`             | `512`     | Atlas texture width, in pixels.                                                                            |
-| `--texture-height <n>`            | `512`     | Atlas texture height, in pixels.                                                                           |
-| `--help`, `-h`                    |           | Print usage.                                                                                                |
+| Flag                             | Default  | Meaning                                                                                                           |
+| -------------------------------- | -------- | ----------------------------------------------------------------------------------------------------------------- |
+| `--font <path>`                  | required | The input `.ttf`/`.otf` file.                                                                                     |
+| `--out <path>`                   | required | Output path with no extension; writes `<out>.png` and `<out>.json`.                                               |
+| `--charset <name\|string\|path>` | `ascii`  | `ascii` for printable ASCII (32-126), a literal string of characters, or a path to a text file to read them from. |
+| `--size <n>`                     | `42`     | The font size, in pixels, the distance field is authored at.                                                      |
+| `--distance-range <n>`           | `4`      | The distance field's encoded range, in pixels.                                                                    |
+| `--texture-width <n>`            | `512`    | Atlas texture width, in pixels.                                                                                   |
+| `--texture-height <n>`           | `512`    | Atlas texture height, in pixels.                                                                                  |
+| `--help`, `-h`                   |          | Print usage.                                                                                                      |
 
 ## Picking a charset
 
@@ -75,11 +78,3 @@ with an error telling you to reduce the charset or increase
 glyphs across multiple pages. Multi-page atlases aren't supported: every
 `FontAtlas` maps to exactly one texture, so a `FontAtlasCache` lookup never
 needs to bind more than one image per font.
-
-## Regenerating after a font or charset change
-
-The output JSON is fully derived from the input font and flags, there's
-nothing to hand-edit. If your game later needs a character the atlas
-doesn't have (a new supported language, an added symbol), re-run the same
-command with an updated `--charset` and commit the new `.png`/`.json` pair;
-there's no incremental update path.
--- a/documentation-site/docs/docs/text/generating-a-font-atlas.md
+++ b/documentation-site/docs/docs/text/generating-a-font-atlas.md
@@ -1,0 +1,85 @@
+---
+sidebar_position: 1
+---
+
+# Generating a Font Atlas
+
+`forge-generate-font-atlas` is an offline command that turns a `.ttf`/`.otf`
+font into an MSDF atlas: `<out>.png` (the atlas texture) and `<out>.json`
+(the glyph metrics, in Forge's own
+[`FontAtlasData`](/Forge/docs/api/interfaces/FontAtlasData) shape). It's a
+build-time step, run once whenever your font or charset changes, not
+something your game runs at runtime.
+
+## Install the generator
+
+The command wraps [`msdf-bmfont-xml`](https://www.npmjs.com/package/msdf-bmfont-xml),
+an **optional peer dependency** of `@forge-game-engine/forge`. A plain
+`npm install @forge-game-engine/forge` does not install it, since most games
+don't need the atlas generator itself, only its output. Install it once,
+as a dev dependency of your own project:
+
+```bash
+npm install --save-dev msdf-bmfont-xml
+```
+
+Running the command without it installed fails fast with a message telling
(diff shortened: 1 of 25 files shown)


documentation-site/docs/docs/text/index.md
```diff
@@ -0,0 +1,58 @@
...
+# Text
+
+The `text` module turns a `.ttf`/`.otf` font into a **multi-channel signed
+distance field (MSDF) atlas**: a small texture plus a metrics file that
+describes every glyph's shape and spacing. An MSDF atlas stays crisp at any
+render size (12px or 400px, zoomed in or out) without regenerating anything,
+because the texture stores each glyph's distance-to-edge rather than a fixed
+raster of pixels.
+
+:::info
```
stormmuller: Useless info, remove it.

documentation-site/docs/docs/text/index.md
```diff
@@ -0,0 +1,58 @@
...
+
+- An offline command, `forge-generate-font-atlas`, that reads a font file
+  and writes out an atlas image plus a
+  [`FontAtlasData`](/Forge/docs/api/interfaces/FontAtlasData)-shaped JSON
+  file, sized to your game's actual `charset`.
+- [`FontAtlasCache`](/Forge/docs/api/classes/FontAtlasCache), which loads
+  that JSON/image pair at runtime into a
+  [`FontAtlas`](/Forge/docs/api/interfaces/FontAtlas), following the same
+  [`AssetCache`](/Forge/docs/api/interfaces/AssetCache) contract as
+  [`ImageCache`](/Forge/docs/api/classes/ImageCache) (see
```
stormmuller: There is no need to reference the image cache here, only Asset Cache.

documentation-site/docs/docs/text/index.md
```diff
@@ -0,0 +1,58 @@
...
+  [Asset Loading](../asset-loading/index.md)).
+
+Guides in this section:
```
stormmuller: This adds not value. The user can see these 2 sections in the nav bar. Remove the "Guides in this section:" blurb from this document.

documentation-site/docs/docs/text/generating-a-font-atlas.md
```diff
@@ -0,0 +1,85 @@
...
+build-time step, run once whenever your font or charset changes, not
+something your game runs at runtime.
+
+## Install the generator
+
+The command wraps [`msdf-bmfont-xml`](https://www.npmjs.com/package/msdf-bmfont-xml),
+an **optional peer dependency** of `@forge-game-engine/forge`. A plain
+`npm install @forge-game-engine/forge` does not install it, since most games
+don't need the atlas generator itself, only its output. Install it once,
+as a dev dependency of your own project:
```
stormmuller: or global dep considering that most games don't need it at runtime.

documentation-site/docs/docs/text/generating-a-font-atlas.md
```diff
@@ -0,0 +1,85 @@
...
+
+```bash
+npm install --save-dev msdf-bmfont-xml
+```
+
+Running the command without it installed fails fast with a message telling
```
stormmuller: This section is useless to a game dev. remove it. The error being descriptive is a concern of the engine maintainer not the read of these docs.

documentation-site/docs/docs/text/generating-a-font-atlas.md
```diff
@@ -0,0 +1,85 @@
...
+you to install it, rather than a raw module-not-found error.
+
+## Generate an atlas
+
+```bash
+npx forge-generate-font-atlas --font my-font.ttf --charset ascii --out assets/fonts/my-font
+```
+
+This writes `assets/fonts/my-font.png` and `assets/fonts/my-font.json`.
+Commit both to your project; [`FontAtlasCache`](./loading-a-font-atlas.md)
```
stormmuller: No need to tell the user to commit.

documentation-site/docs/docs/text/generating-a-font-atlas.md
```diff
@@ -0,0 +1,85 @@
...
+## Only one atlas page is supported
+
+If a charset doesn't fit in the requested texture size, generation fails
+with an error telling you to reduce the charset or increase
+`--texture-width`/`--texture-height`, rather than silently splitting the
+glyphs across multiple pages. Multi-page atlases aren't supported: every
+`FontAtlas` maps to exactly one texture, so a `FontAtlasCache` lookup never
+needs to bind more than one image per font.
+
+## Regenerating after a font or charset change
```
stormmuller: Usless documentation section, remove it

documentation-site/docs/docs/text/loading-a-font-atlas.md
```diff
@@ -0,0 +1,108 @@
+---
+sidebar_position: 2
+---
+
+# Loading a Font Atlas
+
+[`FontAtlasCache`](/Forge/docs/api/classes/FontAtlasCache) loads a
+generated atlas's JSON metrics and PNG texture into a
+[`FontAtlas`](/Forge/docs/api/interfaces/FontAtlas), the same
+`get` / `load` / `getOrLoad` shape as
```
stormmuller: This is not the place to document the asset cache shape. It is already documented elsewhere. Also no need to reference the Image Cache here. it has nothing to do with text.

documentation-site/docs/docs/text/loading-a-font-atlas.md
```diff
@@ -0,0 +1,108 @@
...
+[`ImageCache`](/Forge/docs/api/classes/ImageCache) (see
+[Asset Loading](../asset-loading/index.md) for the shared
+[`AssetCache`](/Forge/docs/api/interfaces/AssetCache) contract).
+
+```ts
+import { FontAtlasCache } from '@forge-game-engine/forge/text';
+
+const fontAtlasCache = new FontAtlasCache();
+const fontAtlas = await fontAtlasCache.getOrLoad('assets/fonts/my-font');
```
stormmuller: It's important to note that this needs to be pointing at the json file. without the extension. 

But when we load any other asset we expect the extension. So this is actually inconsistent and I expect it to be changed.

documentation-site/docs/docs/text/loading-a-font-atlas.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+## Cache keys point at a JSON/image pair
+
+The key you pass is a path with no extension. `getOrLoad('assets/fonts/my-font')`
+fetches `assets/fonts/my-font.json`, then loads whatever image that JSON's
+`atlasImage` field names, resolved relative to the JSON file itself (so the
+`.png` doesn't have to share the exact base name as the key, though
+`forge-generate-font-atlas` always names them to match). As with
+`ImageCache`, the key is an exact string: `'assets/fonts/my-font'` and
```
stormmuller: No need to explain caching behaviour here or mention the image cache. They are documented elsewhere and are out of scope for this document.

documentation-site/docs/docs/text/loading-a-font-atlas.md
```diff
@@ -0,0 +1,108 @@
...
+`'./assets/fonts/my-font'` are cached separately even though they'd resolve
+to the same files.
+
+:::caution
```
stormmuller: Nope, out of scope, remove this section

documentation-site/docs/docs/text/loading-a-font-atlas.md
```diff
@@ -0,0 +1,108 @@
...
+Like `ImageCache`, `getOrLoad` does not deduplicate _in-flight_ loads.
+Calling it for the same key from two places before the first call resolves
+starts two separate fetches. Preload the fonts you need up front (e.g. with
+`Promise.all`) rather than calling `getOrLoad` lazily wherever a font is
+first needed.
+:::
+
+## Malformed or incompatible atlas files fail loudly
```
stormmuller: Remove the fluffy adjectives "loudy". They offer no value to the user and it's just noise.

documentation-site/docs/docs/text/loading-a-font-atlas.md
```diff
@@ -0,0 +1,108 @@
...
+
+`load`/`getOrLoad` reject with a descriptive error if the JSON fails to
```
stormmuller: This is not valuable to the reader, this is the responsibility of us, the maintainers of the engine to provide proper errors. Remove this.

stormmuller:
I was quite unhappy with the general approach to the documentation and need you to update the documentation skill to yield better results in the future. My biggest issues were: 

- The use of fluffy adjectives. Documentation is not fiction, it needs to be clear and concise, the user needs to be able to finds what they need as fast as possible. Documentation should be factual, detailed and succinct.
- Reference to internals, like how we catch certain errors and surface more helpful errors. Or how caching works under the hood. Users care about the interface not the internals.
- Referencing related but unhelpful adjacent parts of the system, For example, when documenting the font cache, it's ok to provide a "for more information see Asset Cache". As the text cache is a type of asset cache. But do not reference "sibling" concepts like the image cache. This adds no value to the reader, it assumes that the reader should be familiar with the image cache. Which is a false assumption. It adds unnecessary cognitive load, for little value. The only reason you added it in is because it was in your context as you were looking for an example on how to implement the new cache. And that is not a good reason to include it in the documentation.

PR #608, opened 2026-08-16
feat(text): add outline and soft-shadow effects, clamped to never overlap a neighboring glyph

stormmuller:
## Summary

Re-attempts Phase 4 of MSDF text rendering (previously implemented in #591/follow-on, then fully reverted in #598) against a corrected root cause, worked out with @stormmuller on #584 before implementing.

**What was wrong before:** every fix in #598's 11 iteration commits (clamping distance-field saturation, fading shadow coverage, bumping the demo atlas's `distanceRange` 4→32) tuned the shader's effect-size clamp against the atlas's own encoded budget only. None of them ever clamped against the actual on-screen gap to a neighboring glyph. Tightly kerned pairs ("il", "ff") can sit close enough that an outline/shadow within the atlas's own budget still visually reaches into the neighbor's ink — and since glyphs are separate alpha-blended draw instances, whichever one is submitted later simply overwrites the earlier one's pixels. That's the "later glyph paints over the earlier one" bug the maintainer diagnosed.

**The fix:** `shapeText` now computes each glyph's own `effectClearance` at shape time — half the gap (in world units) to the tighter of its left/right same-word neighbors, from the same kerning-aware advance walk that already positions glyphs. This is carried per-glyph (not per-entity) through a new `textEffectsInstanceDataSegment`, and `msdf.frag` clamps every effect to `min(atlas budget, this glyph's own clearance)`, converted to screen-pixel-range units *dynamically* via a new `msdf.vert`-forwarded world-per-UV ratio, so the clamp stays correct across camera zoom exactly like `screenPxRange` itself does. No schema change, no quad growth — the existing Phase 1-3 quad geometry (already padded by the atlas's own `distanceRange`) was correct all along; see the corrected-proposal comments on #584 for the full investigation trail, including a retraction of an earlier (wrong) `tilePadding` schema-field idea.

## Related issue(s)

Addresses #584 (Phase 4 of MSDF text rendering)

## Verification

Beyond the checklist below: a real-GPU e2e test (`e2e/specs/text-effects-overlap.spec.ts`) builds a synthetic two-glyph MSDF atlas kerned tightly enough that the *padded quads* overlap by 11 world units while the *ink* stays a real 5 world units apart — reproducing the reported failure mode, not a hand-picked worst case — and proves a `outlineWidth: 50` never paints one glyph's outline color over its neighbor's own ink, while an unaffected third glyph (spaced comfortably wide) still renders its outline at full strength. This is on top of unit tests for the CPU-side `effectClearance` math (`shape-text.test.ts`) and the new instance data segment, since shader-level bugs like this are only visible to a real GPU render (unit tests mock WebGL).

The demo's Liberation Sans atlas was regenerated at `distanceRange: 16` (from `4`) for a larger safe effect budget, and the text demo gained a conservative, documented-safe outline/soft-shadow section.

- [x] `npm test` passes (1163 tests, 165 files)
- [x] Any new/changed public API is exported from the module's `index.ts` (and `/src/index.ts` / `package.json` `exports` — no new module here, `@forge-game-engine/forge/text` already existed)
- [x] Documentation under `/documentation-site/docs/docs` is updated: new `text/text-effects.md` guide (safe range, its two independent limits, and the inherent MSDF glyph-artifact caveats), cross-linked from `rendering-text.md`
- [x] This touches a module with a demo: ran `npm run build`, `documentation-site`'s `npm run typecheck`/`npm run build`, and loaded the demo + new doc page in a real browser (Playwright/Chromium) — screenshots confirmed the outline/glow render distinctly and fit the demo canvas without overlap/clipping

## Changelog

- [x] Added a bullet under `## [Unreleased]` → `#### Added` in `CHANGELOG.md`

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jy6nE1bjrgRsKHWRyK61wL)_


stormmuller:
## Found and fixed a real bug when stress-testing at higher values (`e675b70`)

Per a request to verify this actually works at higher outline/glow values against the real demo font — not just the synthetic single-glyph-pair fixture the original fix's tests used — I built a standalone harness against the real, generated Liberation Sans atlas and pushed `outlineWidth`/`shadowSoftness` up to 24 (the historical target ceiling) on a stress string covering acute-corner glyphs (`W M A V X`) and tight pairs (`ll`, `ff`).

**Found a genuine bug:** the neighbor-gap clamp (`effectClearance` in `shape-text.ts`) measured the gap between adjacent glyphs' *padded quad* edges (`planeBounds`/`atlasBounds`, which include the atlas's baked-in `distanceRange / 2` padding). For a real font at normal atlas-generation settings, that padding is routinely *larger* than the actual space between two ordinarily-spaced letters — so the computed gap was negative (clamped to 0) for nearly every adjacent glyph pair, not just deliberately tight kerned ones. Confirmed empirically: `"AB"` at `outlineWidth: 4` rendered with **no outline at all**, while `"A B"` (space-separated, i.e. different words) rendered fine. This would have made the effect essentially non-functional on any real multi-letter word.

**Fix:** derive each glyph's actual ink edges (subtracting the atlas's own padding, computed from `distanceRange` and this glyph's own atlas-pixels-to-world-units ratio - both already in `FontAtlasData`, no schema change) and measure the neighbor gap from ink edges instead of padded quad edges. Full derivation and reasoning in the commit message and `assignEffectClearances`'s doc comment.

**Re-verified against the real atlas** with the same stress string at `outlineWidth`/`shadowSoftness` 4/10/16/24 and `shadowSoftness` 4/12/24:
- Ordinary adjacent letters (the large majority of any real string) now render a full, clean outline/glow at every tested value.
- Genuinely tight pairs (`ff`, `AV` in this specific font - they sit close together by design) still correctly suppress toward a thin/near-zero outline rather than let two letters' rings merge - the deliberate, documented conservative trade-off, now correctly scoped to pairs that are actually tight instead of every adjacent pair.
- No corner artifacts or counter-closing observed on `W M A V X O` at any tested value with this font/string.
- No overlap between distinct letters at any tested value, including 24.

Added a unit test constructing a deliberately extreme synthetic kerning value to confirm genuinely-touching ink still clamps to 0 (`shape-text.test.ts`), and corrected the two existing tests that had been (unknowingly) asserting the buggy padded-quad-based behavior. All unit tests, e2e tests, and the full verification suite pass.

---

stormmuller:
## Found and fixed the real cause of "outlineWidth doesn't work, I expected a large border" (`58c3faa`)

You were right - this was a real bug, not a perception issue. Root-caused it against the live demo rather than just re-checking the math.

**What was wrong:** `assignEffectClearances` in `shape-text.ts` was *halving* the ink gap between two same-word neighbors, giving each glyph only half the gap as its own safe budget. For ordinary running text (not just tight kerned pairs like "il"/"ff"), that halved budget is tiny - e.g. on the demo's own "Forge Engine" at its caption size, the tightest letter pair ("rg" in "Forge") only had ~0.5 world units of clamped headroom. Any `outlineWidth`/`shadowSoftness` above that got silently clamped to the same sliver, so cranking the value from 2 to 10 to 40 visibly did *nothing* - confirmed by rendering the real demo at increasing values and diffing the screenshots pixel-for-pixel; they were identical. An isolated single glyph (no neighbor to clamp against) rendered a correctly large, clearly-scaling border at the same values, which is what pointed at the neighbor clamp specifically rather than the shader math.

**The fix:** the halving was solving for the wrong invariant. The actual thing that must never happen is one glyph's effect painting over a *neighbor's own ink* - it's completely fine for two glyphs' outline/shadow *layers* to overlap each other in the gap between them (they're typically the same color, and blending order only matters right at an actual glyph's own boundary, which is still protected). So each glyph can safely claim the **full** gap up to - but never past - its neighbor's ink, not half of it. This doubles the usable budget without reintroducing the original overlap bug.

**Also found while re-verifying:** the e2e test's `contestedWorldX` (meant to sample a point inside glyph A's own ink) was actually computing glyph A's *padded quad* edge, landing the sample point inside neighbor glyph B's ink instead - so the "never paints outline over a neighbor's ink" test was passing without actually exercising the clamp. Fixed the derivation and added `e2e/specs/text-effects-overlap.spec.ts`'s new "still visibly grows with outlineWidth, not stuck at a sliver" test as a permanent regression guard for this specific failure mode.

**Verification beyond the existing suite:**
- Re-ran the full real-GPU e2e suite (22/22 passing) plus the new regression test.
- Rebuilt the demo's dist output and added a large "hero" example (`FORGE` at a bigger on-screen size) to the live text demo, proving a properly bold, clearly-scaling outline + soft glow - screenshot posted to the conversation, and the demo itself now ships this example permanently so it's not just a one-off proof.
- Corrected the small caption-scale showcase's own values, which (I found in the process) were themselves already slightly outside the *true* safe budget for their own sample text - reduced to a value that's actually safe.
- Updated `text-effects.md` to stop overselling a flat "4 screen-pixel-range units is safe" claim - the neighbor budget is a *percentage* of the text's on-screen size, not a fixed pixel constant, so small caption text will only ever support a thin effect. That's now stated explicitly instead of left for someone to discover the hard way.

All local checks (`check-types`, `test` - 1165 passing, `lint`, `cspell`, `check-exports`, docs-site `typecheck`/`build`) pass; CI should confirm shortly.

---

